PR #11, opened 2015-05-20
Remove space in extension list options as invalid

infosec-au:
The options of dirs3arch include a space when specifying comma delimited extensions. The script needs extensions passed in without comma's. Just to help people that don't realise at first, I've removed the space in the help output.


maurosoria:
Thanks dude, merged

PR #12, opened 2015-05-25
Fixed a bug when opening wordlists with broken utf-8 and updated the …

jonaslejon:
…complete list with wordpress plugins (old 42k, new 51k).

Fixes the following error:

UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 3912-3913: invalid continuation byte


maurosoria:
Thanks bro :D

PR #13, opened 2015-05-28
Update Requester.py

dgaavl:
Nothing important, just eliminate the last "?" char that actually is appended in all the url requests


maurosoria:
Merged :D

PR #14, opened 2015-05-28
Update FuzzerDictionary.py

dgaavl:
When reading the dictionary file any word that doesnt ends with char "/" or doesnt contains "%EXT%" is added two times:
1) just as is
2) with a "/" appended

So, if the dictionary file is like this:

# DictionaryFile
web
index.%EXT%

and the command is called this way:
$ ./dirs3arch -w DictionaryFile -e "php,html" -u url

the fuzzed dictionary will be:

web
web/
index.php
index.html

The result is the same we'll get with this dictionary file:

# DictionaryFile
web
web/
index.%EXT%

The reason for this change is that, sometimes, web servers return a 404 when the directory resource is hit but the url resource without "/" returns a differente code what it usually means that it contains more resources...

It's not complicated to create a new dictionary file for this kind of webs but I prefer dirs3arch doing the work for me...

### DIFF
--- a/lib/core/FuzzerDictionary.py
+++ b/lib/core/FuzzerDictionary.py
@@ -62,7 +62,10 @@ class FuzzerDictionary(object):
                 for extension in self._extensions:
                     self.entries.append(urllib.parse.quote(line.replace('%EXT%', extension)))
             else:
-                self.entries.append(urllib.parse.quote(line))
+                if line.endswith('/'): self.entries.append(urllib.parse.quote(line))
+                else:
+                    self.entries.append(urllib.parse.quote(line))
+                    self.entries.append(urllib.parse.quote(line)+'/')
         if lowercase == True:
             self.entries = list(oset([entry.lower() for entry in self.entries]))
 


maurosoria:
Check my comment in this issue:
https://github.com/maurosoria/dirs3arch/issues/10

Would that feature fit your needs?

The problem with this patch, It's that won't fit my needs, because it'd just duplicate a lots of lines of the dictionary (something that made me hate dirbuster in the past... and that's why dirs3arch exists). I like in this way because it will test only the files that i've specified in the dictionary.

But, you're not the first person who want compatibility with fuzzdb/dirbuster dictionaries, so I'd be adding a new argument switch for this mode in this commit (and a new line in default.conf, soy you'd be able to set this mode as default behaviour) , so everyone will be happy.

Do you agree with this? Any other suggestion?

dgaavl:
Sure,
that solution would be perfect.

Thank you very much.
David.

2015-05-29 0:55 GMT+02:00 Mauro Soria <notifications@github.com>:

PR #15, opened 2015-05-28
Update Controller.py

dgaavl:
Modifies paths that could be considered recursive by dirs3arch:

# Web server resources:

www.url.com/web/ - 404
www.url.com/web - 302
www.url.com/web/help/ - 404
www.url.com/web/help - 302
www.url.com/web/help/index.html - 200

# Dictionary file
web/
web
help
help/
index.%EXT%

$ ./dirs3arch -u www.url.com -r -e "html"

Before change: Recursive are only those paths that ends with ("/")

www.url.com/web - 302

After change: Recursive are those paths that are not files (don't contains a dot) and are not just "/" chars.

www.url.com/web - 302
www.url.com/web/help - 302
www.url.com/web/help/index.html - 200

### DIFF
--- a/lib/controller/Controller.py
+++ b/lib/controller/Controller.py
@@ -203,16 +203,21 @@ class Controller(object):
     def addDirectory(self, path):
         if self.recursive == False:
             return False
-        if path.endswith('/'):
-            if path in [directory + '/' for directory in self.excludeSubdirs]:
-                return False
-            if self.currentDirectory == '':
-                self.directories.put(path)
-            else:
-                self.directories.put('{0}{1}'.format(self.currentDirectory, path))
-            return True
-        else:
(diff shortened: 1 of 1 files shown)


maurosoria:
Why this change? I don't understand the purpose of this patch. Directories can be dotted too.

Furthermore, some php CMSs, force you to use routes like this:
http://www.example.com/cms/index.php/users/

In this cases, you have a route that is a file, so this patch won't work.

dgaavl:
Sorry, you're right.

Best regards,
David.


2015-05-29 1:01 GMT+02:00 Mauro Soria <notifications@github.com>:

maurosoria:
No problem. If you have another suggestion, feel free to open an issue :+1:

PR #38, opened 2016-08-24
Added .well-known/

k2l8m11n2:
https://eff.org/dnt-policy
`This file will always
be posted via HTTPS at https://example-domain.com/.well-known/dnt-policy.txt
to indicate this fact.`


maurosoria:
Thanks !

Issue #40, opened 2016-09-15
--ip has no effect


r0p0s3c:
Hi,

In [Requester.py](https://github.com/maurosoria/dirsearch/blob/master/lib/connection/Requester.py#L119), `requests` is passed the URL, so specifying an IP address in `--ip` has no effect. The best method to cover both when IP is supplied and when it is resolved is to call `requests` with the IP (either supplied or resolved) and specify the host part from the URL in the `Host` header.

I will send in a PR to fix this shortly.

PR #41, opened 2016-09-15
Fix to implement --ip functionality

r0p0s3c:
As per #40, this correctly implements functionality to "override" the IP useful in, for example, virtual-host based searching

### DIFF
--- a/lib/connection/Requester.py
+++ b/lib/connection/Requester.py
@@ -103,7 +103,7 @@ class Requester(object):
             try:
                 if self.proxy is not None:
                     proxy = {"https" : self.proxy, "http" : self.proxy}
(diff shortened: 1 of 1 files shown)


maurosoria:
Hello k2l8m11n2 , thanks for the contribution,

Would you please change your format string syntax? You are using the classic syntax with %, and I'm currently using the new "{}".format syntax in the whole project. Check this

https://docs.python.org/3.1/library/string.html#format-examples

Please, I'd like to keep the code consistent.

k2l8m11n2:
Hey, the patch was actually written by r0p0s3c and I only approved the commit.

maurosoria:
Thanks r0p0s3c

PR #53, opened 2017-02-13
Added option to force requests by hostname

liamosaur:
By default, dirsearch queries by IP, with the hostname in the Host header.

This is fine for most webservers, but some hosts (like CDNs) can get picky about this and reject requests that are made by IP address.

I added a boolean -b / --request-by-hostname argument that will override the default behaviour and instead query by hostname

### DIFF
--- a/lib/connection/Requester.py
+++ b/lib/connection/Requester.py
@@ -41,7 +41,7 @@ class Requester(object):
         }
 
     def __init__(self, url, cookie=None, useragent=None, maxPool=1, maxRetries=5, timeout=30, ip=None, proxy=None,
-                 redirect=False):
+                 redirect=False, queryByHosename=False):
         # if no backslash, append one
         if not url.endswith('/'):
             url = url + '/'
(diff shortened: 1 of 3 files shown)


maurosoria:
Many thanks for the contribution, never had that problem before, but any improvement is welcomed

thanks again

Jack-ban:
@maurosoria @liamosaur Hello! Please tell me how to use `-b` / `--request-by-hostname` argument.
I also encounter this error:
```bash
 _|. _ _  _  _  _ _|_    v0.3.8
(_||| _) (/_(_|| (_| )

Extensions: CHANGELOG.md | HTTP method: get | Threads: 10 | Wordlist size: 6099

Error Log: /root/dirsearch/logs/errors-19-09-08_11-44-48.log

Target: https://cpay.westcoal.com.cn/

SSL Error connecting to server. Try the -b flag to connect by hostname

Task Completed
```
First, I think my target is not under CDN's protection. I can surf the websites by IP.
Second I have been try to add `-b`, but it's doesn't work. So I'm not sure whether use it in wrong way or not.
as follows 
```bash
python3 dirsearch.py -u "https://www.target.com/" -b  -e *
```
or 
```bash
python3 dirsearch.py -u "https://(target's IP)/" -b "https://www.target.com/"  -e *
```
it is wrong?

PR #54, opened 2017-02-21
Update BaseReport.py

SUHAR1K:
In the beginning I wanted to correct an error in the url, but it seemed to me that it is better to leave it in primary view that will display the structure of the resource. I do not know whether the error is in Linux.


maurosoria:
Thanks dude!!!

PR #57, opened 2017-03-29
add delay option

redshark1802:
allows setting a delay in seconds between requests: ``--delay n``


maurosoria:
Thanks for the contribution!!!! very helpful

PR #59, opened 2017-06-01
suppress empty output patch

liamosaur:
I've recently come across a bunch of websites that annoyingly return empty 200 responses (see screenshot)
screen shot 2017-06-01 at 15 00 03

I added a `--suppress-empty` flag that stops 0 byte responses from being logged to the console


maurosoria:
Hi liaosaur! Great job! Many thanks

PR #65, opened 2017-09-01
fix: 'NameError: name 'thirdparty' is not defined'

eur0pa:


LockeLamora:
amazing, works fine for me

PR #66, opened 2017-09-01
Allow recursive search with third party wordlists

LockeLamora:
dirsearch's own dictionary specifically had '/' appended to words it wanted to test as directories, but other wordlists such as from dirbuster don't. This meant that recursion wouldn't work as Controller.addDirectory would specifically test for the presence of '/' in a successful search term to judge if it's a directory to be recursively searched. 

This fix will, as the terms are read in from the dictionary, append a '/' if one isn't present (and also if a '.' isn't present), but also keep the raw word for submission also. so for example if 'install' is in the dictionary, 'install' and 'install/' will both be tested.

The downside of this is that with the default dictionary since 'install' and 'install/' are already separately defined this may add overhead as the same term will be searched multiple times, but as it's a small dictionary i don't think it should be too much of a problem.  A possible mitigation for this would be to wrap the current fix in a check that ensures we aren't using the default dirsearch dictionary.

### DIFF
--- a/lib/controller/Controller.py
+++ b/lib/controller/Controller.py
@@ -334,10 +334,10 @@ class Controller(object):
     def addDirectory(self, path):
         if not self.recursive:
             return False
-        if not (".") in path:
+        if path.endswith('/'):
             if path in [directory + '/' for directory in self.excludeSubdirs]:
                 return False
-            self.directories.put(self.currentDirectory + path + '/')
+            self.directories.put(self.currentDirectory + path)
             return True
         else:
             return False
--- a/lib/controller/Controller.py
+++ b/lib/controller/Controller.py
@@ -334,10 +334,10 @@ class Controller(object):
     def addDirectory(self, path):
         if not self.recursive:
             return False
-        if path.endswith('/'):
+        if not (".") in path:
             if path in [directory + '/' for directory in self.excludeSubdirs]:
                 return False
-            self.directories.put(self.currentDirectory + path)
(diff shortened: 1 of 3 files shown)


maurosoria:
Hi LockeMora, thanks for the provided fix

I have checked your patch, and I'm hesitating to merging. I actually think that, as you pointed out, there is too much overhead in the install' and 'install/' problem. I think the next is enough:

```
entry + "." + ext
entry + "/"
```

In that case, I think that modifying lib.core.Dictionary.generate() method, then forced extension is used, in order to append a "/" character (lines 72-73).

What is your opinion?

LockeLamora:
ah ok so it -f is used then we assume it's something like a dirbuster list and append both extensions and trailing slashes? that'd work if we also do a check for an existing trailing slash (just to prevent a double slash)  and if we update the help to specify that -f is useful for most custom wordlists?

yeah that'd be cool in my opinion

maurosoria:
Check the patch I've submitted. Tell me if it works for you and your personal opinion.

LockeLamora:
I'll have a look in a few hours :) it has to take into account existing extension search terms and empty terms (I've put a revised solution up if you need to see some) as i've noticed these exist in other wordlists, there was an issue where an empty line in the dirb common list would find '/' as a positive result then search it all again due to recursiveness for example. The solution i have up worked fine for a hackthebox.eu machine

I'll have a look at yours when I'm more able I'd like to see how you did it too :)

maurosoria:
No response form LockeLamora.

PR #71, opened 2017-10-26
Generate dictionary without dot bug + Filter by Size

adon90:
I have noticed that, when trying to look for no extension files, it automatically adds a dot at the end.
I have fixed that dictionary creation not to add a dot in that case.
I have also included an option in order to filter responses by size, as there are cases when the response is a 200 and is actually a "Not Found" with a fixed size.


maurosoria:
Hi adon90, 

First of all, my apologies for no answering before. 

I've never accepted de pull request because I was planning to add this functionality myself. The reason is because the way I want to implement it is the following:
* The possilibity to use multiple filters, not just once, in combination with the status code. For example: --filter-size-1 100 --filter-status-1 200. (--filter-size-<i> --filter-status-<i>, where <i> is a number from 1 to 20)
* Using bytes is simple not enough, I was thinking using the same format of the gui (for example, 10B, 10K, 10M). So, if you specify 1kb, you're including from 1024 bytes to 2043 bytes.
* Using ranges, for example, 300-380, could be a possible value, 100B-2KB too.

Haven't done yet, but with some help it might be easier

adon90:
Hello Mauro, sorry for my late response, I didn't want to relay you until I
verify the --filter-size using ranges works smoothly.
Your tool is really overshadowing dirb and other tools, many people use
them and that's awesome, you probably you it given the number of stars your
github has haha!
I attach 2 pictures showing how it is working.
In my github last commit you can see the changes, since you don't accepts
pull requests :P. But it's ok, I'm sure my code sucks since I'm not a
programmer and you can probably to it cleaner:
https://github.com/adon90/dirsearch/commit/ce3e540bf8f37bc284c4c0b953a6f61f2cdcd083
Regads
btw, if you point me as contributor in your github as "adon90" i'd be so
happy and would be such an honor for me, I always keep suggesting using
your tool to anyone :P!

2018-03-08 21:18 GMT+01:00 Mauro Soria <notifications@github.com>:

PR #93, opened 2018-02-11
Update dicc.txt - Add configuration.php.zip

shahril96:
Some sysadmins backed up their configuration in a zip file.


maurosoria:
great!!!

PR #98, opened 2018-05-24
Add hint to connect by hostname for sites that require SNI

liamosaur:
By default, dirsearch makes requests by IP address, not hostname.

For SSL/TLS sites that host multiple domains on the same IP address, this means the requests made by IP address by dirsearch will fail, as the SSL/TLS connection requires a hostname.

Currently dirsearch fails in this situation without any hints as to what the problem is, and can require some troubleshooting to figure out what's going on:
screen shot 2018-05-24 at 13 55 48

This change detects the specific exception caused by this condition, and adds a hint to try re-running with the -b flag, that tells dirsearch to make requests by hostname, not IP:

screen shot 2018-05-24 at 14 00 20


maurosoria:
Nice one!

PR #113, opened 2018-10-05
Update dicc.txt

act1on3:
add:
`app.js` and `server.js` for checking JS server code. Sometimes admins are using directory for static files as a JS server code (misconfiguration problem).
Link: [blog.blackfan.ru](https://blog.blackfan.ru/2018/01/pda-test.yandex.ru-file-reading.html)


maurosoria:
Thanks bro!

PR #114, opened 2018-10-13
Added status code as output

babutzc:
I added the status code which dirsearch is looking for. The first time i used it, i didn't know which HTTP code's it was bruteforcing. 

There for i added it in the source code.


maurosoria:
The tool doesn't look for specific status codes, but tries to discover content using differents tests.

The commit it is also invalid because it doesn't represent what the tools is doing.

PR #122, opened 2019-01-23
Recurse on directory redirects

Sjord:
When in recursive mode, if the response is a redirect to a directory, schedule it to scan later.

The redirect needs to be resolved relative to the current URL to make relative redirects work. We only add it to self.directories when it is a subdirectory, to avoid endless loops when two subdirectories redirect to one another.

This is pretty experimental. I have tested it on a test server, but not in production yet.

### DIFF
--- a/lib/controller/Controller.py
+++ b/lib/controller/Controller.py
@@ -20,6 +20,7 @@ import gc
 import os
 import sys
 import time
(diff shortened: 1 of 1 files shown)


maurosoria:
Thanks for this patch! Haven't checked but it seems good!

PR #132, opened 2019-03-15
Discovery update - Spring boot actuators

tkisason:
Added a few lines that should help with discovery of spring boot actuators...

See here why this is important:

https://www.veracode.com/blog/research/exploiting-spring-boot-actuators


maurosoria:
Thanks bro!

PR #158, opened 2019-07-21
Ran python Black formatter to improve code readability/maintenance

hangmansROP:
This might not be that useful but I saw from previous issues that you're looking to improve readability/maintainability via something like AutoPEP8 etc. I normally use the [Python Black](https://github.com/python/black) plugin/formatter which is great for consistency and also has plugins for various IDE's/editors. I've ran the formatter on the repo and the result is in this PR. Let me know what you think!


jsfan:
Not sure this will do much for the actual style issues. Black is great but it mainly does things that go beyond PEP8. Right now, the issue are probably the PEP8 violations not that the style is too inconsistent across files.

PR #168, opened 2019-09-30
Bugfix for --scan-subdirs+--recursive, added checks for duplicate scans, added request count, fixed types, extended readme, merged with #158 (formatting) & #166 (excluded text/regex)

Damian89:
Hi, 
when scan-subdir was used the "calculation" of currentDir and basePath were wrong, this resulted in wrong requests when also recursive scanning was used. This should be fixed now.

Additionally now no more duplicate scans happen (especially when scanner found the same folder the user added additionally via scan-subdir).

I also added the request count which basically is:
wordlistsize (extension amount calculated also)*scan-subdirs.

- Additionall added stuff from #166 
- Formating like in #158 
- Fixed some typos
- Added some additional examples to readme

Best,
Damian


maurosoria:
hi Damian89

Thanks for the PRs. 

The code has change since this PR. I would accept the PRs if you submit separately, so is easier to test/rollback.

maurosoria:
hello @Damian89 

I've just merged some of the changes, thanks for your contribution!

Regards,
Mauro

PR #171, opened 2019-10-02
[Doc] Add feature : Exclude By Text / Regex

jenaye:


jenaye:
thx 4 add this  @trololomgwtf it'll be useful

PR #181, opened 2019-10-22
Make it possible to use as command

tkmru:
This pull request make it possible to use as `dirsearch` command. It will be more useful.
I think it would be bad to change the usage suddenly, and `dirsearch.py` is still there. I can also delete this file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,23 @@
 reports/
 logs/
 
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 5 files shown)


tkmru:
@maurosoria I made it available as a dirsearch command, what do you think?

maurosoria:
Hi tkmru, thanks for the contribution

I've been accepting in the last days multiple pull requests. Even I am grateful for that, I have noticed that someone introduced a deadlock in the code and now it doesn't finish the task as it should.

I will take a time to try to find what happened (it had happened before and it was a hard to fix it) and after that I will check this. I don't want to move forward without fixing this issue.

Thanks for your help and I hope you understand.

Regards,
Mauro

tkmru:
Hi, Maruo.
Thank you for your sincere reply. A deadlock bug is hard. I wish you good luck.

.gitignore
```diff
@@ -1,6 +1,23 @@
 reports/
 logs/
 
+# Distribution / packaging
```
Lucifer733, line 4: I fixed this issue please give me my money

shelld3v: This PR seems old with a lot of conflicts. BTW, this is IDP, not IBP

Issue #144, opened 2019-05-16
Need help for PEP8


maurosoria:
Hi community!

I don't have much time lately, and I'd like to start applying this convention. So, I'm opening this issue looking for people who have time to do it.

If anyone is there, and you have experience with this convention, I'd really appreciate your help.

https://www.python.org/dev/peps/pep-0008/

I started writing this tool many years ago and I was used to write using lowerCamelCase Convention (hadn't written enough python code yet and didn't know so many people would use this tool really...) but know that I see that this piece of code might have some future, I think that using a convention it's necessary for the community driven development.

Anyone there?

Damian89:
Yeah, I can help you here - step by step, I like and need dirsearch, so expect some prs in the near future which contain just renames etc

Damian89:
Btw. who is this "Daniel89" in the readme ? :P

maurosoria:
Would you forgive me ? :c

Damian89:
of course ;)

ftk-sostupid:
Maybe we can use https://github.com/hhatto/autopep8 ?
I can help too if needed. ：）

PR #183, opened 2019-10-28
Some initial PEP8 cleanup

jsfan:
This is just the low-hanging fruit. More to come.

Sending the PR now to make you aware.

This partially address #144.


jsfan:
This still doesn't remove all warnings in `pycodestyle` or `flake8`. However, the question is if you want to follow PEP8 in all aspects. Many projects disagree e.g. with the line length limit and disable that warning.

jsfan:
Once all warnings are gone or ignored, it would make sense to set up a Travis pipeline to enforce them for future patches. That would then also allow for adding some unit testing later.

maurosoria:
@jsfan @shelld3v 


Can we do this now?

shelld3v:
No, the code has changed so much, so we can't know if this pull request can create any problem. We can use this as an example of how to cleanup pep8 someday. I will do something like this in the future, safely, but not now. I think we can close this, thank you @jsfan, we appreciate your help and will use this as a doc in further updating!!!

jsfan:
If I find the time, I'll go through the code as it is now and redo the cleanup.

shelld3v:
Thank you ❤️

PR #187, opened 2019-11-05
Add option to use default extension list

### LINKED ISSUE #146: [Feature-Request] --extensions-list - option for providing a list of extensions
Extract from the help message:

```
  Mandatory:
    -u URL, --url=URL   URL target
    -L URLLIST, --url-list=URLLIST
                        URL list target
    -e EXTENSIONS, --extensions=EXTENSIONS
                        Extension list separated by comma (Example: php,asp)
```

Having an `--extensions-list` would be nice to be able to use a long list of extension rather than listing them all by hand.

Example of extension dict: fuzzdb/discovery/predictable-filepaths/filename-dirname-bruteforce/Extensions.Common.txt

mzfr:
Fix #146

I have added an option called `--extension-list` that stores value `True` or `False`. So if this option is provided then a predefine small list of the extension will be used in place of extension. I haven't used the wordlist suggested by @noraj i.e `: fuzzdb/discovery/predictable-filepaths/filename-dirname-bruteforce/Extensions.Common.txt` because this file had like 800 extension and using that many extension as the default would slow down the process.

Now the hardcoding of the common extension might seem odd so we can either add an option to read those default extensions from the configuration file or we can also have the option of supporting the extension file. But I think the support for extension file should be added for the existing argument `-e or --extension` instead of this new `--extension-list`. 

If anything else looks bad in this PR then please let me know.

### DIFF
--- a/lib/core/ArgumentParser.py
+++ b/lib/core/ArgumentParser.py
@@ -58,9 +58,12 @@ class ArgumentParser(object):
         else:
             self.urlList = [options.url]
 
-        if options.extensions == None:
-            print('No extension specified. You must specify at least one extension')
+        if not options.extensions and not options.extensions_list:
+            print('No extension specified. You must specify at least one extension or try using default extension list.')
             exit(0)
+
+        if not options.extensions and options.extensions_list:
+            options.extensions = "html,php,tar.gz,txt,xml,zip,jpg,png,jpeg"
 
         with File(options.wordlist) as wordlist:
             if not wordlist.exists():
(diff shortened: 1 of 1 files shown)


noraj:
@mzfr Thx for your PR. My idea in #146 was that you should provides a file to `--extension-list` like for `-L URLLIST` rather than hard-coding anything.

lib/core/ArgumentParser.py
```diff
@@ -58,10 +58,13 @@ def __init__(self, script_path):
...
             self.urlList = [options.url]
 
-        if options.extensions == None:
-            print('No extension specified. You must specify at least one extension')
+        if not options.extensions and not options.extensions_list:
+            print('No extension specified. You must specify at least one extension or try using default extension list.')
             exit(0)
 
+        if not options.extensions and options.extensions_list:
+            options.extensions = "html,php,tar.gz,txt,xml,zip,jpg,png,jpeg"
```
noraj: Read extensions from a FILE rather than hardcoding them, see https://github.com/maurosoria/dirsearch/blob/26a9e240bd8de93108f9ac1c400ea49f8634e4fa/lib/core/ArgumentParser.py#L36-L59

noraj:
@maurosoria You merged the PR with hardcoded extensions ?

noraj:
@maurosoria bump

PR #206, opened 2019-12-31
Enable to use multiple dictionaries by '-w' option.

zrquan:
When I want to use multiple dictionaries by '-w' option, I can separate them with commas.


maurosoria:
Thanks for this commit!

PR #210, opened 2020-01-13
Multiple dirsearch improvements

V-Rico:
Here are some improvements that can be useful for different situations:

- Option to remove dot from extension when forcing (--nd, example%EXT% instead of example.%EXT%)
- Options to display only items with response length from range (--min & --max)
- Option to whitelist response codes (-i 200,500)
- Option to remove output from console (-q, keeps output to files)
- Option to add custom suffixes to filenames without dots (--suff .BAK,.old, example.%EXT%%SUFFIX%)
- Multiple output improvements

_TODO:_
_make suffix option working only for found files_

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,3 +182,5 @@ Special thanks for these people.
 - k2l8m11n2
 - vlohacks
 - r0p0s3c
+- V-Rico
+- russtone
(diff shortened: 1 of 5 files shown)


V-Rico:
Any update?

maurosoria:
Many thanks ! Love your work. Sorry for being late in approving this

Regards,
Mauro

PR #250, opened 2020-07-02
adding more command-line history files, SSH, Bash and VIM config files

kazet:


kazet:
Wow, that was fast :)

PR #275, opened 2020-08-31
Fix #274

### LINKED ISSUE #274: Remove urljoin in CLIOutput.py
I think we should use another method to parse the path in **CLIOutput.py** rather than `urllib.parse.urljoin`. The reason is that `urllib.parse.urljoin` can not print the original path with `./`, `../` or `////`:

```
>>> urllib.parse.urljoin('/', 'path')
'/path'
>>> urllib.parse.urljoin('/', './path')
'/path'
>>> urllib.parse.urljoin('/', '../path')
'path'
>>> urllib.parse.urljoin('/', '///path')
'/path'
```

This is why that when you add `../../../../../../../etc/passwd` into your wordlist, the output will become:

```
200 -  20B   -  etc/passwd
```
**Correct output**
```
200 -  20B   -  /../../../../../../../etc/passwd
```

shelld3v:
Fix #274. I have tested VERY carefully and found that my solution can parse everything the same as `urllib.parse.urljoin`. The only difference is that `posixpath.join` won't edit anything in the path even though there are `..`, `.` or how much `/` in the path:

```python
>>> from posixpath import join as urljoin
>>> urljoin('/', 'test')
'/test'
>>> urljoin('/', 'test/')
'/test/'
>>> urljoin('/', 'test/blah')
'/test/blah'
>>> urljoin('/foo', 'test/blah')
'/foo/test/blah'
>>> urljoin('/foo/bar', 'test/blah')
'/foo/bar/test/blah'
>>> urljoin('/', '../../../../../../etc/passwd')
'/../../../../../../etc/passwd'
>>> urljoin('/', './admin')
'/./admin'
>>> urljoin('/', '/////something')
'/////something'
>>> urljoin('https://google.com', 'path')
'https://google.com/path'
```

### DIFF
--- a/lib/output/CLIOutput.py
+++ b/lib/output/CLIOutput.py
@@ -22,6 +22,7 @@ import threading
 import time
 
 import urllib.parse
+from posixpath import join as urljoin
 
 from lib.utils.FileUtils import *
 from lib.utils.TerminalSize import get_terminal_size
@@ -101,11 +102,11 @@ class CLIOutput(object):
                 contentLength = FileUtils.sizeHuman(size)
 
             if self.basePath is None:
-                showPath = urllib.parse.urljoin("/", path)
+                showPath = urljoin("/", path)
 
             else:
-                showPath = urllib.parse.urljoin("/", self.basePath)
-                showPath = urllib.parse.urljoin(showPath, path)
+                showPath = urljoin("/", self.basePath)
+                showPath = urljoin(showPath, path)
             message = "[{0}] {1} - {2} - {3}".format(
                 time.strftime("%H:%M:%S"), status, contentLength.rjust(6, " "), showPath
             )
--- a/lib/output/CLIOutput.py
+++ b/lib/output/CLIOutput.py
@@ -21,7 +21,6 @@ import sys
 import threading
 import time
 
-import urllib.parse
 from posixpath import join as urljoin
 
 from lib.utils.FileUtils import *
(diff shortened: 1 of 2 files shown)


shelld3v:
You can test everything locally first

shelld3v:
Any update?

shelld3v:
Please give me an update!

shelld3v:
Merge this plz

shelld3v:
Why no one merge this?

shelld3v:
ping

shelld3v:
ping

shelld3v:
ping

maurosoria:
Hello @shelld3v,

I didn't merge it because I don't know if posixpath is available on Windows. I googled but couldn't check. Do you know?

Sorry for not answering.

shelld3v:
Checked and it's worked:

```python
Python 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from posixpath import join as urljoin
>>> urljoin('/', 'test')
'/test'
>>>
```

maurosoria:
Let's see how it behaves.

From my experience this kind of changes are problematic, but I like risks =)

Have you seen this ? https://github.com/maurosoria/robologin-dev I'd like you to take a look, at least the code

shelld3v:
Hmm, not bad! I will take a look ❤️

shelld3v:
Are you sure it's a good thing to create a username+password brute-forcer when there are so many tools like that: Burp Intruder, ffuf, ...?

PR #278, opened 2020-08-31
Allow any HTTP method rather than just GET, POST, HEAD (and fixed some stuff)

shelld3v:


maurosoria:
Hello @shelld3v 

The PR has a dangerous eval function. I cannot accept this PR, I am sorry.

Regards,
Mauro

maurosoria (changes requested):
look for method self.session.request(self.httpmethod, url, **kwargs)
requests.request(method, url, **kwargs)[source]
https://requests.readthedocs.io/en/latest/api/

shelld3v:
Edited! Is that correct?

shelld3v:
>The PR has a dangerous eval function

We have already checked the HTTP method in **Controller.py** if it is valid or not so I don't think it's dangerous

PR #284, opened 2020-09-01
Detect 429 Too Many Requests

shelld3v:
Added 429 Too Many Requests detector

- I haven't tested locally yet, so you should probably check it before applying to the code
- I haven't used colors in the output message, you can do it if you want


maurosoria (changes requested):
Using the Output class is important. No prints are allowed.

shelld3v:
OK, I also don't think that we should add this functionality in **Requester.py**

PR #303, opened 2020-09-04
Refresh default wordlist (remove duplicates, add some stuff, delete some useless endpoints and sort everything)

shelld3v:
From the default dicc.txt, I just sort everything, then remove, add some stuff (actually most of the changes are adding new endpoints)


shelld3v:
If you can't view changes, then [here](https://gist.github.com/shelld3v/090334c2f3e40c84d6cd721573c88577)

shelld3v:
```python
dicc = list(dict.fromkeys(dicc))
dicc.sort()
```
=> Current wordlist

maurosoria:
Actually, one file for wordlists hasn't been convincing from long time..

maybe use some file for generate the wordlist?

for example

```
[admin]
path='{BASE_PATH}/admin-files.txt'

[juicyfiles]
path='{BASE_PATH}/juicyfiles.txt'

[uploads]
path='{BASE_PATH}/uploads.txt'
forces=True
`
```

You got my point? wouldn't that be cool? hackers would make their own tempaltes.

shelld3v:
I have thought about it for a long time LMAO!!! 

But, there are several problems: many hackers may don't know to edit the template. And if they actually want to brute-force each server with each wordlist, they can use the `-w` flag. Most of the hackers I know, @dawgyg for example, enough smart to make many wordlists for many situations, use the `-w` flag to run that wordlist. Create the template file to work with many wordlists will very complex. 

I think that keep one file for wordlist is a really good thing, since people who actually do not good at assets discovery still can simply work with dirsearch, don't need to setup stuff and choosing the correct wordlist. And good hackers who actually want to make many wordlists to brute-force for each situation can add files into the wordlist directory and use `-w` to work with it. 

Even about me, I am still using one wordlist to brute-force most of the time (of course I have a lot of other wordlists to brute-force if I actually know what the server is running with: IIS, ...).

The most important problem here is it's too COMPLEX, it will make a lot of trouble for hackers. (I think most of the hackers using dirsearch probably don't know that `default.conf` is editable, since there is no documentation for that lol, I will create a doc for dirsearch in the future)

maurosoria:
> I have thought about it for a long time LMAO!!!
> 
> But, there are several problems: many hackers may don't know to edit the template. And if they actually want to brute-force each server with each wordlist, they can use the `-w` flag. Most of the hackers I know, @dawgyg for example, enough smart to make many wordlists for many services, use the `-w` flag to run that wordlist. Create a template file to work with many wordlists will very complex.

I do not want to remove the actual behaviour of the -w argument. The thing about the default wordlist, is many hacker (no bug hunters, but pentesters) find it really useful for their daily work, because is a small, compact, but get you enough info to start.

In my case, for example, I am also used as using different wordlists (obviously the default one is not enough, just a starting point) but always the same patterns; if it is a windows, I use asp,aspx in lowercase, if it is a wordpress I use a wordpress plugin wordlist, etc.

What I said previously is something i want to change: it would continue to support the previous Dictionaries (default and forced extensions) but with a template, at least we can organize the line in separate files (because now maintaining that is a nightmare) and if someone wants to PR a new path with juicy info, there is a wordlist for that (specified in the template), a url of a CVE-, a backend path or some api endpoints, .... etc (you get the point, just only look how de dicc.txt commits look like, impossible to trace).

And with the same idea, you can have separate templates, passing as an argument (let's just call it --wordlist-template) so a person can craft a template just with some lines and create his own templates based in other wordlists with his own workflows (remember order of the wordlist it's also important, and that now with a single file it's difficult to fix...)

Tell me what's your opinion

shelld3v:
Yeah, it may be good to add the templates option, but must still keep the additional wordlist process workflow of dirsearch, since I don't want to make more problems for the hackers who actually don't want to work with templates setup

PR #311, opened 2020-09-04
add opendir detection

futex:
I used this simple dirty trick to detect open dir. I find it very usefull in CTF.

dirsearch

### DIFF
--- a/lib/output/CLIOutput.py
+++ b/lib/output/CLIOutput.py
@@ -111,7 +111,10 @@ class CLIOutput(object):
             )
 
             if status == 200:
-                message = Fore.GREEN + message + Style.RESET_ALL
+                if "index of" in str(response.body).lower():
+                    message = Fore.GREEN + message + Style.RESET_ALL + " OpenDir"
+                else:
+                    message = Fore.GREEN + message + Style.RESET_ALL
 
             elif status == 403:
                 message = Fore.BLUE + message + Style.RESET_ALL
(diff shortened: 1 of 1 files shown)


shelld3v:
This detector sucks. `Index of` in the body content does not mean that page is in directory listing mode, there will be a lot of false positives here. Beside, the message `OpenDir` sucks too, it should be something like ` -> Directory listing`, ... And I think this function is not useful at all, since the users can open the page on their browsers themself.

futex:
sorry i would opened an issue to know if someone find it useful, not open a PR for this Q&D poc

shelld3v:
Next time, create an issue about your suggestion and I will create a pull request for it 👍

its0x08:
Hey @shelld3v if I'm being honest that something I wanted to have for long now.
This implementation is not the best but the feature is definitely needed IMO!
Also `dirsearch` shouldn't `brute force` if there is `dir listing`, but instead I think
would be better to `parse` the links on the `dir listing` and check them for `size`, `status code`, etc..
Directories inside a `dir listing` should be `checked` for `dir listing` and if not should be `brute forced`.
Doing this manually is a pain!
Hope I expressed my self clearly 😁

shelld3v:
Great idea, but I will have to do a lot of things to can make it happen

its0x08:
Sooner of later it will have to be done I guess, so would be better to start splitting it in smaller tasks and do it by stages not all at once.

PR #317, opened 2020-09-06
New feature: Exclude extensions (close #268)

### LINKED ISSUE #268: [Suggestion] Adding "exclude extensions" option
`dirsearch` should add a new option called "exclude extensions". This allows the user to select the extensions that he doesn't want to brute force with. For example, while hacking on Yahoo, I found that some of Yahoo servers didn't allow access to  `.jsp` files. Even the file existed or not (`blahthisdoesnotexist.jsp`), the server will always return a 403 error. So the "exclude extensions" option will not send any request which is `.jsp` extension, reduce spam outputs. The workflow is simple: remove all endpoints in the list that end with `.jsp`, then continue by brute-forcing with that endpoints list. 

Also, I want to develop dirsearch, can I join the dirsearch's developers team and help dirsearch grows the functionalities in directory brute-forcing? I will very careful when doing that, by testing with my fork first, then let you take a look at the update and apply if it's accepted by you.

Thank you for your attention

shelld3v:
For more information: #268

### DIFF
--- a/lib/controller/Controller.py
+++ b/lib/controller/Controller.py
@@ -120,7 +120,7 @@ class Controller(object):
 
         self.dictionary = Dictionary(self.arguments.wordlist, self.arguments.extensions, self.arguments.suffixes,
                                      self.arguments.lowercase, self.arguments.forceExtensions,
-                                     self.arguments.noDotExtensions)
+                                     self.arguments.noDotExtensions, self.arguments.excludeExtensions)
 
         self.printConfig()
         self.errorLog = None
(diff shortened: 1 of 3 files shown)


shelld3v:
Close #268

maurosoria (changes requested):
you shoud user a filter for thi, at the same time you convert this to list (to only pass every line one time)

shelld3v:
I can't get your point, I just updated a little bit in my PR, is that the thing you are talking about?

shelld3v:
If you are trying to say something like:

```python
for res in list(oset(result)):
```
Then you should know that `remove()` will only *remove the first instance of a value in a list*:

```python
>>> test = [1, 3, 5, 7, 3]
>>> test.remove(3)
>>> test
[1, 5, 7, 3]
```

PR #321, opened 2020-09-09
New feature: Exclude extensions (close #268)

### LINKED ISSUE #268: [Suggestion] Adding "exclude extensions" option

shelld3v:


shelld3v:
Dude, plz do something good with the issues! Please set the labels, read all of them, ... If you are going to ignore all the issues, then at least set labels into it so I will work with them. Also please check out my issues, there are some stuff I want to ask or suggest you

maurosoria:
Hello @shelld3v,

I look for the PRs in my free time. Also the same as the last ticket, a set does no preserve the order that user wants.

It's not olny a matter of performance

shelld3v:
>Also the same as the last ticket, a set does no preserve the order that user wants.

I am using `oset` in my code 🤦

shelld3v:
Also, can you recheck #275

PR #322, opened 2020-09-09
Remove duplicates in the URL list

shelld3v:
For more info about why I use `list(set())`: https://towardsdatascience.com/3-easy-and-noninvasive-ways-to-instantly-boost-your-python-code-performance-b21cd823f37e

### DIFF
--- a/lib/controller/Controller.py
+++ b/lib/controller/Controller.py
@@ -141,7 +141,7 @@ class Controller(object):
             )
 
         try:
-            for url in list(set(self.arguments.urlList)):
+            for url in list(dict.fromkeys(self.arguments.urlList)):
                 try:
                     gc.collect()
                     self.reportManager = ReportManager()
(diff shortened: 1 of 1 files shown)


maurosoria:
Hello @shelld3v ,

A normal SET does not preserve the order, that's why an ordered set is better.

shelld3v:
Updated! 

Also, performance is SUPER IMPORTANT. It's the reason why so many hackers turned from dirsearch to ffuf. The speed is one of the things I am going to focus on it now before release dirsearch 0.4.0. Totally, I have used ffuf and it's actually SO FAST that impressed me a lot than dirsearch. The point here is that most of the hackers now are having the main wordlist with 50000-200000 endpoints, and if you run it with ffuf, it's so much better than dirsearch. @mayonaise, who has stopped using dirsearch and turn to ffuf said that he did it because of the performance, so I don't want it to happen anymore.

PR #329, opened 2020-09-10
Fix #304

### LINKED ISSUE #304: Bug in the wordlist generator
Hi,

When I ran dirsearch with the -e flag which for extensions, I received this output:

```python
Extensions:  | HTTP method: GET | Suffixes: html, htm, php, jsp, asp, js, json ...
```
It showed that there is no extension and there are several suffixes. The special thing here is that my wordlist itself was 8132 endpoints, then after rendered with several extensions, it supposed to be around 10000 endpoints and 10000 requests. But from the output:

```python
Wordlist size: 8052 | Request count: 8052 (+recursive)
```

Why is this happening, is it a bug? Can you explain this? My wordlist is something like this:

```
admin
index.%EXT%
login
api/
server.php
```
This is making me feel regret for updating dirsearch, it looked like using dirsearch v3.0.8 maybe better! Before everything gets fixed, I will use `ffuf` for sure 👌

shelld3v:
dirsearch created a loop to add all the extensions into the entry:

```python
for extension in self._extensions:
```

But you accidentally edited the entry after the first time:

```python
if self._noDotExtensions:
    line = reextdot.sub(extension, line)

line = reext.sub(extension, line)
```

That removed the `%EXT%` and caused the bug 🤕

### DIFF
--- a/lib/core/Dictionary.py
+++ b/lib/core/Dictionary.py
@@ -98,11 +98,11 @@ class Dictionary(object):
                 if '%ext%' in line.lower():
                     for extension in self._extensions:
                         if self._noDotExtensions:
-                            line = reextdot.sub(extension, line)
+                            entry = reextdot.sub(extension, line)
 
-                        line = reext.sub(extension, line)
+                        entry = reext.sub(extension, line)
 
-                        quote = self.quote(line)
+                        quote = self.quote(entry)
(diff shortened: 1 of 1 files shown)


maurosoria:
I didn't code the nodotextensions, that's why checking PRs is difficult :S

PR #334, opened 2020-09-11
Boost wordlist generator

shelld3v:
#### Can `dict.fromkeys()` preserve the order?
Yes

```python
>>> from oset import *
>>>
>>> alist = ['l', 'b', 'd', 'a', '1', 'd', '1', '2']
>>> list(oset(alist))
['l', 'b', 'd', 'a', '1', '2']
>>> list(dict.fromkeys(alist))
['l', 'b', 'd', 'a', '1', '2']
```

#### Why we should change from `oset()` to `dict.fromkeys()`?
Boost dirsearch, that's all.

**Comparison:**
```python
>>> from oset import *
[0.00971442184448242 secs]
>>> 
>>> alist = []
>>> for x in range(500000):
...     alist.append(x)
...
>>> 
>>> test1 = list(oset(alist))
[1.609781770706177 secs]
>>> test2 = list(dict.fromkeys(alist))
[0.1053035259246826 secs]
```

**Results:**
- `oset()`: 0.0097 secs + 1.6097 secs = 1.6194 seconds  
=> 1.6 seconds
- `dict.fromkeys()`: 0.1053 seconds 
=> 0.1 seconds

That is when we use a 500,000 endpoints wordlist. Haven't tested with a 1 million endpoints wordlist 🙄

### DIFF
--- a/lib/core/Dictionary.py
+++ b/lib/core/Dictionary.py
@@ -87,7 +87,7 @@ class Dictionary(object):
 
         # Enable to use multiple dictionaries at once
         for dictFile in self.dictionaryFiles:
-            for line in dictFile.getLines():
+            for line in list(dict.fromkeys(dictFile.getLines())):
 
                 # Skip comments
                 if line.lstrip().startswith("#"):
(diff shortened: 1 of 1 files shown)


shelld3v:
Please also merge #322 if you agree with my solution.

PR #340, opened 2020-09-11
Speed up the code

shelld3v:
What I have learned today:
*"Pythonic code is not only more readable but also faster, another example: 'if variable:' is faster than the un-idiomatic 'if variable == True:'"*

`if test:` and `if test is not None:` are the same. It only causes an error if `test` actually returned `False`, which matches with `if test:` but not `if test is not None:`. Unfortunately, I didn't see anything like that in your code, it only returns `None` or an `str` object. 

```python
>>> test = None
>>> if test:
...     print(test)
...
>>> test = 'blah'
>>> if test:
...     print(test)
...
blah
```

### DIFF
--- a/lib/connection/Requester.py
+++ b/lib/connection/Requester.py
@@ -79,7 +79,7 @@ class Requester(object):
         self.host = parsed.netloc.split(":")[0]
 
         # resolve DNS to decrease overhead
-        if ip is not None:
+        if ip:
             self.ip = ip
         else:
             try:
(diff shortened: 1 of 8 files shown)


shelld3v:
I have also finished a test:

```python
>>> timer()
0.00742030143737793
>>> timer()
0.007399559020996094
>>> timer()
0.007498502731323242
```

```python
>>> timer2()
0.00518488883972168
>>> timer2()
0.005187511444091797
>>> timer2()
0.0051953792572021484
```
--------------------------------------------
```python
>>> timer()
0.019031763076782227
>>> timer()
0.01875782012939453
>>> timer()
0.018575191497802734
```

```python
>>> timer2()
0.0128173828125
>>> timer2()
0.013179540634155273
>>> timer2()
0.012797832489013672
```
----------------------------------------------
```python
>>> timer()
0.037598371505737305
>>> timer()
0.03699016571044922
>>> timer()
0.03717923164367676
```

```python
>>> timer2()
0.02648186683654785
>>> timer2()
0.02586960792541504
>>> timer2()
0.02572941780090332
```

PR #341, opened 2020-09-12
Speed up the wordlist generator by "avoiding dot"

shelld3v:
Read from a blog post! 

**First solution:**

```python
def timer():
     start_time = time.time()
     for x in range(500000):
         test.append(x)
     print(time.time() - start_time)
``` 
Results:

```python
>>> timer()
0.07306313514709473
>>> timer()
0.08280205726623535
>>> timer()
0.07710886001586914
```
**Second solution:**

```python
 def timer():
     start_time = time.time()
     append = test.append
     for x in range(500000):
          append(x)
     print(time.time() - start_time)
```
Results:

```python
>>> timer()
0.047969818115234375
>>> timer()
0.04876255989074707
>>> timer()
0.04903578758239746
```

It faster by twice

### DIFF
--- a/lib/core/Dictionary.py
+++ b/lib/core/Dictionary.py
@@ -85,6 +85,7 @@ class Dictionary(object):
         reext = re.compile('\%ext\%', re.IGNORECASE)
         reextdot = re.compile('\.\%ext\%', re.IGNORECASE)
         result = []
+        append = result.append
 
         # Enable to use multiple dictionaries at once
         for dictFile in self.dictionaryFiles:
@@ -103,7 +104,7 @@ class Dictionary(object):
                         entry = reext.sub(extension, line)
 
                         quote = self.quote(entry)
-                        result.append(quote)
+                        append(quote)
 
                 # If forced extensions is used and the path is not a directory ... (terminated by /)
                 # process line like a forced extension.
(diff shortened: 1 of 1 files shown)


maurosoria:
Hello @shelld3v,

I have to check this with patience. The thing, it's even the time you describe appears to be faster, the code readability also matters.

```def timeit(method):
    def timed(*args, **kw):
        ts = time.time()
        result = method(*args, **kw)
        te = time.time()
        if 'log_time' in kw:
            name = kw.get('log_name', method.__name__.upper())
            kw['log_time'][name] = int((te - ts) * 1000)
        else:
            print('%r  %2.2f ms' %
                  (method.__name__, (te - ts) * 1000))
        return result
    return timed```

then you do somethins like this:
```@timeit
def somefunc():
```

shelld3v:
Can you explain more about *the code readability*!

maurosoria:
> Can you explain more about _the code readability_!

Of course! But maybe if you google the term **'code readibility vs performance'**  you will find more detailed explanations:
For example: https://stackoverflow.com/questions/183201/should-a-developer-aim-for-readability-or-performance-first

Obviously, renaming the variable "append" to another thing may help, but ...

Can you pass me the blogspot link please? I'd like to read it to understand better.

Finally, are we a sure a regular python list is the most suited solution if you are looking for performance? I wonder about that too.

PR #343, opened 2020-09-12
Update Dictionary.py

shelld3v:
Before:

```
.info.jsp
backup.jsp
admin.jsp
phpmyadmin.php.jsp
phpmyadmin.inc.php.jsp
aspnet.asp.jsp
js-files.js.jsp
```

After:

```
.info.jsp
backup.jsp
admin.jsp
phpmyadmin.php
phpmyadmin.inc.php
aspnet.asp
js-files.js
```

### DIFF
--- a/lib/core/Dictionary.py
+++ b/lib/core/Dictionary.py
@@ -107,17 +107,17 @@ class Dictionary(object):
 
                 # If forced extensions is used and the path is not a directory ... (terminated by /)
                 # process line like a forced extension.
-                elif self._forcedExtensions and not line.rstrip().endswith("/"):
+                elif self._forcedExtensions and not line.rstrip().endswith("/") and not "." in line.rstrip()[1:]:
                     quoted = self.quote(line)
 
                     for extension in self._extensions:
                         # Why? check https://github.com/maurosoria/dirsearch/issues/70
-                        if extension.strip() == '' or ('.' in quoted and not quoted.startswith('.')):
+                        if extension.strip() == "":
                             result.append(quoted)
                         else:
-                            result.append(quoted + ('' if self._noDotExtensions else '.') + extension)
+                            result.append(quoted + ("" if self._noDotExtensions else ".") + extension)
 
-                    if quoted.strip() not in ['']:
+                    if len(quoted.strip()):
                         result.append(quoted)
                         result.append(quoted + "/")
 
--- a/lib/core/Dictionary.py
+++ b/lib/core/Dictionary.py
@@ -112,7 +112,7 @@ class Dictionary(object):
 
                     for extension in self._extensions:
                         # Why? check https://github.com/maurosoria/dirsearch/issues/70
-                        if extension.strip() == '':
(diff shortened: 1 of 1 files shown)


maurosoria:
Hello @shelld3v ,

First, thanks for your work. I cannot really show you how happy I am in a github comment =)

I am trying to understand:

* Why someone would put that lines in a forced extension wordlist?
* Why this approach if now the %NOFORCE% keyword is available?
* What if the a lines has a dot, but it's not because an exension? That would cause unexpected behaviour.
* Can you explain me the intended workflow ?

shelld3v:
Usually, a "dirsearch" wordlist will be looked like this:

```
login
server/
admin.%EXT%
```
Different from others:

```
admin.html
admin.php
login
server/
```

So if there is any endpoint in the wordlist that is using a dynamic extension: `file.extension`, then probably that extension is unique and only for that endpoint. Example:

```
admin.%EXT% (every extension can be here)
login.%EXT% (every extension can be here)
index.%EXT% (every extension can be here)
```
and

```
phpmyadmin.php (PHP My Admin is a PHP based manager, there won't be something like phpmyadmin.asp or phpmyadmin.php.asp)
ASPnet.asp (An ASP server's file)
Trace.axd (A default IIS file)
```

>What if the a lines has a dot, but it's not because an exension?

Do you mean `admin.`? If then, why that hacker doesn't use `admin.%EXT%` or `admin` + force extensions? Absolutely he wants to test with ONLY `admin.`, no extension after it!

```
login
admin.
admin.%EXT%
test
```
===> 

```
login
admin.
admin.php
admin.html
admin.js
admin.asp
test
```

>Why this approach if now the %NOFORCE% keyword is available?

This is a good question! But now, most of the hackers have already had a wordlist. And the wordlist size usually around 10000-100000 entries. Adding `%NOFORCE%` after all of the endpoints that shouldn't be forced is impossible, so this solution may be good!

Of course, I don't know will there be any problem here and if you don't agree with my solution, you can close this!

Even what, if you are not gonna merge this, then you still should do something like this:

```
admin.php
login.php
```
===>

```
admin.php
admin.html
admin.asp
login.php
login.html
login.asp
```

Rather than:

```
admin.php
admin.php.html
admin.php.asp
login.php
login.php.html
login.php.asp
```

maurosoria:
Actually, I think is useful too. But what happens if there is a wordlist and a folder like this:

somepage.com/folder/com.somepackage.somethng

This will only try only 1 posibility. I know the case it's weird, but as I previously said this change can cause unexpected behaviour (miss some cases)

shelld3v:
I have searched Google for those types of files and found that those endpoints should not be forced with extensions 👍

shelld3v:
Hi, any update? com.somepackage.something is a package file, like a plugin. Not a thing to force for sure!

shelld3v:
In the future, I may create a argument to force all the entries no matter it ends with an extension or not!

shelld3v:
I will make a better solution soon!

Issue #298, opened 2020-09-03
-H flag does not work


shelld3v:
This is my dirsearch command:
```basb
dirsearch.py -u http://localhost:4444 -e php -H google:google
```
And from my server:
```bash
$ nc -lvp 4444
listening on [any] 4444 ...
connect to [127.0.0.1] from localhost [127.0.0.1] 19951
GET / HTTP/1.1
User-agent: Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/28.0.1468.0 Safari/537.36
Accept-Encoding: identity
Accept: */*
Connection: keep-alive
Accept-Language: en-us
Keep-Alive: 300
Cache-Control: max-age=0
Host: localhost:4444
```

I didn't see the header `google:google` in the incoming request! Is it a bug?

shelld3v:
I think the problem is from **Requester.py**, from the code:
```python
    def __init__(
        self,
        url,
        cookie=None,
        useragent=None,
        maxPool=1,
        maxRetries=5,
        delay=0,
        timeout=30,
        ip=None,
        proxy=None,
        proxylist=None,
        redirect=False,
        requestByHostname=False,
        httpmethod="get",
    ):
``` 
I don't see the *headers* variable - the headers from the user input!, The only headers I saw is the default headers:
```python
headers = {
        "User-agent": "Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/28.0.1468.0 Safari/537.36",
        "Accept-Language": "en-us",
        "Accept-Encoding": "identity",
        "Accept": "*/*",
        "Keep-Alive": "300",
        "Connection": "keep-alive",
        "Cache-Control": "max-age=0",
    }
```
You should add the user headers variable and then combine it with the default `headers`, before sending the request

shelld3v:
Actually, I just saw this line:
```python
                        self.requester.setHeader(key, value)
```
With this:
```python
    def setHeader(self, header, content):
        self.headers[header] = content
```
I actually don't know what is going wrong here? 🤔

shelld3v:
This is a bug dude, give it a label!

maurosoria:
this is still unpatched? Tell me and I will check it.

shelld3v:
Yeah, it's still there!

shelld3v:
Actually, it's my fault

PR #368, opened 2020-09-14
Remove --quiet-mode

shelld3v:
Since we haven't had a workflow for it yet, we should remove it so people won't use that "empty" flag


shelld3v:
Hi, can you also check #298! If you can't reproduce the bug, tell me!

PR #370, opened 2020-09-14
Update from BeautifulSoup to BeautifulSoup4

shelld3v:


shelld3v:
I think that you should add BeautifulSoup4 into the `thirdparty` folder. The gitable source is here: https://github.com/waylan/beautifulsoup

PR #371, opened 2020-09-14
Update Controller.py

shelld3v:


shelld3v:
Dude, can you add flags to ALL the issues? I know that you don't have too much free time to resolve it but give it some flags may help me to manage it easier (I will try to solve all the issues). Give them flags will be also looked friendlier with the reporters

PR #394, opened 2020-09-17
Bug Fix

serhattsnmz:
Fix the following error:

```
  File "./dirsearch.py", line 26, in <module>
    from lib.core import ArgumentParser
  File "/home/serhat/Desktop/dirsearch/lib/core/__init__.py", line 2, in <module>
    from .Dictionary import *
  File "/home/serhat/Desktop/dirsearch/lib/core/Dictionary.py", line 30
    def __init__(self, paths, extensions, suffixes=None, prefixes=None, lowercase=False, uppercase=False, forcedExtensions=False, noDotExtensions=False, excludeExtensions):
                 ^
SyntaxError: non-default argument follows default argument`
``

### DIFF
--- a/lib/core/Dictionary.py
+++ b/lib/core/Dictionary.py
@@ -27,7 +27,7 @@ from lib.utils.FileUtils import File
 class Dictionary(object):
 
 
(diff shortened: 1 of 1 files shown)


shelld3v:
LOL Sorry, I used to do that but thought it is useless so removed it

shelld3v (approved):

PR #410, opened 2020-09-19
update docker tag to python:3-alpine and reduced image to 80MB (was 1GB)

jsav0:
The base `python:3-alpine` image tag is only 42.7MB whereas the current default `python:3` tag is 882MB

The current dockerfile when built comes out to a big 1.03GB.

I suggest building with alpine instead and it reduces the image to 80.4MB. 

Tested and works.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM python:3
-MAINTAINER greenmind.sec@gmail.com
-RUN apt-get update -y
-RUN apt-get install python3-pip -y
+
+FROM python:3-alpine
+LABEL maintainer="wfnintr@null.net"
 WORKDIR /root
-ADD . .
-WORKDIR /root/
-RUN chmod +x dirsearch.py
-ENTRYPOINT ["/root/dirsearch.py"]
+RUN apk add --no-cache --virtual .depends git 
+RUN git clone https://github.com/maurosoria/dirsearch.git
+RUN apk del .depends
+WORKDIR /root/dirsearch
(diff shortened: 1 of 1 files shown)


shelld3v:
Can you tell me what is the difference between `python:3` and `python:3-alpine`?

jsav0:
> Can you tell me what is the difference between `python:3` and `python:3-alpine`?

1. `python:3` is based off of debian buster and glibc and therefore inherently huge. [source](https://github.com/docker-library/repo-info/blob/master/repos/python/local/3.5-buster.md)
2. `alpine:3` is based off of alpine (8MB) and musl and is inherently small. [source](https://github.com/docker-library/repo-info/blob/master/repos/python/local/3-alpine3.12.md) 

Oftentimes, unless you are debian+glibc dependent, you don't need 800MB of system overhead just to run a simple application. Thus alpine+musl is often preferred for docker images and containers.

jsav0:
The other thing I changed in my commit of the Dockerfile is how the source is added. 
Before, it was necessary to clone the repo *and then* build the image. Now the source is being cloned **inside** of the container. This eliminates the need to clone the source locally first to build.
You can just download the Dockerfile by itself and run it
```
wget https://raw.githubusercontent.com/maurosoria/dirsearch/master/Dockerfile
docker build . -t dirsearch
```
Or you can host the image somewhere, and pull and exectute the container directly, for example
```
docker run --rm -it wfnintr/dirsearch
```

shelld3v:
For your interest, I don't know ANYTHING about Docker, even just a small information about how to use it!

Issue #151, opened 2019-06-20
Couldn't resolve DNS on windows


SaltNego:
````
C:\Users\Negoowen\dirsearch>python3 dirsearch.py -u 'http://123.206.87.240:8002/flagphp/' -e php

 _|. _ _  _  _  _ _|_    v0.3.8
(_||| _) (/_(_|| (_| )

Extensions: php | Threads: 10 | Wordlist size: 5918

Error Log: C:\Users\Negoowen\dirsearch\logs\errors-19-06-20_15-34-29.log

Target: 'http://123.206.87.240:8002/flagphp/'

Couldn't resolve DNS

Task Completed
````

zzy2210:
yet,I have this problem too。Have you solved this problem?

maurosoria:
Can you provide more info? Only in windows with IP:PORT format?

DustinTheGreat:
> Can you provide more info? Only in windows with IP:PORT format?

Can confirm it works on Linux

shelld3v:
The problem here is that in Windows, `"stuff"` will become `stuff` but `'stuff'` will still be `'stuff'`, `'` won't be removed. Try `python3 dirsearch.py -u "http://123.206.87.240:8002/flagphp/" -e php` and see the magic (close the issue too)

shelld3v:
Again, this is not an application issue, it's a human issue! Please close this!

Be1nna:
same problem(Linux)

PR #426, opened 2020-09-27
New feature: Capital wordlist  +  Several updates

shelld3v:
Since we just released dirsearch 0.4.0 and everyone hasn't update it yet, please wait 2-3 days before merging this because we can't know if there is any bug in this code update! If there is really a bug, we can prevent from affecting too many dirsearch users

==> Don't merge this, please wait 2-3 days first

### DIFF
--- a/default.conf
+++ b/default.conf
@@ -18,6 +18,7 @@ autosave-report-format = plain
 wordlist = db/dicc.txt
 lowercase = False
 uppercase = False
+capitalization = False
 force-extensions = False
 no-dot-extensions = False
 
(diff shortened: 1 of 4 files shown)


shelld3v:
Oh just forgot, congrats for the release!!! I'm happy to see that! Also, there are a few things I want to say:
- If you see any issue that suggests a new feature but you don't agree with that, you can add a `won't fix` label
- Please close #151

shelld3v:
Hi, I have confirmed that there is no bug in my code (tested locally)!

PR #438, opened 2020-09-29
New feature: No extension

shelld3v:
This will be useful when you are brute-forcing API or something like that


shelld3v:
Any problem?

PR #445, opened 2020-09-30
Update dicc.txt

shelld3v:


shelld3v:
I deleted some endpoints ended with `.php` and `.html` since they have already had the `.%EXT%`, so the extension will be identified by the user

PR #465, opened 2020-10-05
Update Controller.py

shelld3v:


maurosoria:
Hello @shelld3v ,

Wouldn't it be better to pass the noDelay as a requester attribute?

shelld3v:
I don't think so!

shelld3v:
Can you please close #151!

shelld3v:
Any update? 🌚

PR #494, opened 2020-10-17
Update README.md

ansh052008:
Update README.md


shelld3v:
Hi, the real name of this tool is *dirsearch*, without any uppercase character. That is why I didn't change from *"dirsearch"* to *"Dirsearch"* in that line

PR #446, opened 2020-09-30
PEP8 Compliance

jsfan:
Fixes all PEP8 violations except for line length and wildcard imports.

PR #500, opened 2020-10-17
Snake case

### LINKED ISSUE #144: Need help for PEP8

jsfan:
Following on from #446 this PR transforms some camelCased names to snake_case.

This does **not** close #144 because there are many more occurrences of camelCase which should really be snake_case.


jsfan:
it probably makes no sense to clean up all variable and method names in one PR because that would lead to a very large patch which would be impossible to review.

@shelld3v Does this look good to you so far? Feel free to modify in my repo again.

shelld3v:
Why? I don't understand why you do this, most of variables I created when updating dirsearch are something like camelCased

jsfan:
https://www.python.org/dev/peps/pep-0008/#id45

jsfan:
It's not something `flake8` checks but it is part of PEP8.

shelld3v:
I still don't think it's a good idea

jsfan:
You don't have to go with it. The whole point of PEP8 though is that there are shared guidelines. So, if you were to find a contributor who is used to Python guidelines, they'd likely use snake case just as a Java coder would likely use camel case. Right now, the code base has not fixed style and my conclusion would be to go with what is generally adopted in the language it is written in. But, once again, it's your code base, so I'm only making a suggestion.

maurosoria (approved):

jsfan:
@maurosoria I've resolved the conflicts.

shelld3v:
I actually don't like this PR, but I'm not the owner

maurosoria:
Hello @shelld3v and @jsfan 

I wanted to do this changes by myself, but I never have the opportunity (because it has to be done progressively). What @jsfan is pointing out is totally reasonably: if this is not done, PEP8 makes no sense.

Thanks @jsfan and @shelld3v !

PR #509, opened 2020-10-18
Disable Insecure HTTPS warning

shelld3v:
Description
---------------

Since 2017, `urllib3` not a part of `requests` anymore. That is why my update for the `requests` module created the `InsecureRequestWarning` message


shelld3v:
Solve #512

PR #511, opened 2020-10-18
Check the scheme of the URL

shelld3v:
Description
---------------

Full check the protocol

### DIFF
--- a/lib/connection/requester.py
+++ b/lib/connection/requester.py
@@ -65,9 +65,12 @@ class Requester(object):
 
         self.protocol = parsed.scheme
 
-        # if not protocol specified, set http by default
-        if self.protocol not in ["https", "http"]:
+        # If no protocol specified, set http by default
+        if not self.protocol:
             self.protocol = "http"
+        # If protocol is not supported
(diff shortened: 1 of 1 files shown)


maurosoria:
Hello @shelld3v,

I remember that back in the days urlparse would behaves unexpectedly if you don't provide the scheme. That's why I force the procotol before and after.

Are you sure this PR would work correctly?

shelld3v:
Worked really good on my side:

```python
>>> import urllib.parse
>>> urllib.parse.urlparse("google.com")
ParseResult(scheme='', netloc='', path='google.com', params='', query='', fragment='')
>>>
```

maurosoria:
> Worked really good on my side:
> 
> ```python
> >>> import urllib.parse
> >>> urllib.parse.urlparse("google.com")
> ParseResult(scheme='', netloc='', path='google.com', params='', query='', fragment='')
> >>>
> ```

But look the ParseResult, the path should be the netloc. Probably we should use something like normalize_url

shelld3v:
What about now?

PR #517, opened 2020-10-20
Auto connect by hostname if "SSL Error connecting to server" occurs

shelld3v:
Description
---------------

>SSL Error connecting to server. Try the -b flag to connect by hostname

Now this message will no longer appear since we will automatically "connect by hostname" rather than ask the user for that

### DIFF
--- a/lib/connection/requester.py
+++ b/lib/connection/requester.py
@@ -182,11 +182,8 @@ class Requester(object):
                 raise RequestException({"message": "Too many redirects: {0}".format(e)})
 
             except requests.exceptions.SSLError:
-                raise RequestException(
-                    {
-                        "message": "SSL Error connecting to server. Try the -b flag to connect by hostname"
-                    }
-                )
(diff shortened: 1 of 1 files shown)


maurosoria:
Hello @shelld3v, how you doing? 

I have a small doubt about what happens if the continue in the except block could cause and endless loop.

I've thought about solving this in the past, and it's not so trivial, because you can misplace a CDN rejection (the most common scenario in which you would use -b argument) and a real SSL connection error.

I think that the requester or the scanner should do some scanning in order to check if the -b argument should be used. That can be done using some techniques, if you are interested. 

But my biggest concern is the endless loop.

Great contributions as always, thanks!

shelld3v:
No, after the except block, the `self.url` will be changed to hostname. So the except block won't happen again, since the problem has been solved. FYI, I have already tested my patch in my local machine. 

Also waiting for your responses in other PRs (by me and others :)

PR #522, opened 2020-10-24
Set the default recursion level to infinity

shelld3v:
Description
---------------

Set the default recursion level to infinity


shelld3v:
Recently just know that there is a `lstrip()` that can remove character in the left of the word lol

PR #538, opened 2020-10-31
Important bug fix

shelld3v:
Description
---------------

Don't know why but I remember I have added this in my PR


shelld3v:
Hi, this is a really important bug fix, and you are merging it really slow! Can you give me a way to contact you whenever that is a huge problem in dirsearch?

PR #545, opened 2020-11-03
Update Dockerfile

random-robbie:
Description
---------------

Why this PR? What will it do?

added requests as when running it was missing urllib3

Don't forget :)
---------------

 * Add your name to README.md and CONTRIBUTERS.md
 * If this is a new feature, then please add some additional information about it to `CHANGELOG.md`

Thank you very much for helping dirsearch become a better tool!!! :heart:


shelld3v:
No, this is not important! The `requests` package has already been implemented in the `thirdparty` folder

maurosoria:
Thanks @random-robbie ! 

@shelld3v it make senses that the docker image fetch dependencies in the Dockerfile and not from thirdparty.

PR #556, opened 2020-11-08
Remove useless --ip option

shelld3v:
Description
---------------

The user can --request-by-hostname rather than find the hidden IP address and use with --ip

### DIFF
--- a/lib/connection/requester.py
+++ b/lib/connection/requester.py
@@ -75,7 +75,7 @@ class Requester(object):
         self.host = parsed.netloc.split(":")[0]
 
         # resolve DNS to decrease overhead
-        if not self.requestByHostname:
+        if not requestByHostname:
             try:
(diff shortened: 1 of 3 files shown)


maurosoria:
I use this argument a lot (it's specially useful if there is a hidden virtual host)

shelld3v:
Hi, I know the use case of this option. But I mean now, it's no longer important, since the user can connect directly to the IP address and edit the host header. And if there is a WAF, `--request-by-hostname` is the solution.

maurosoria:
Hello @shelld3v , how you doing?

It's more clear to me to put the entire url with the proper vhost, and then specify the ip address. Then, proper reports and settings will be set (and not reports named like ip addresses).

Other interesting thing, it's very trivial to modify dirsearch to brute force vhosts....

PR #557, opened 2020-11-08
Shitty pep8

shelld3v:
Description
---------------

### DIFF
--- a/lib/connection/requester.py
+++ b/lib/connection/requester.py
@@ -187,17 +187,17 @@ class Requester(object):
                     {"message": "Error with the proxy: {0}".format(e)}
                 )
 
-            except requests.exceptions.ConnectionError as e:
+            except requests.exceptions.ConnectionError:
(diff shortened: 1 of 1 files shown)


jsfan:
If you think it's bad that PEP8 doesn'tlet you have unused variables, never try writing Go. Go won't compile if you have unused variables. ;)

shelld3v:
I know that because I used to learn Golang. I mean that shit rule in Go, it's horrible right?

jsfan:
I love it. I can't stand code bases full of cruft. I've spent a lot of time in my career figuring out code only to find out later that it wasn't actually used. Not allowing unused variables is a good start.

shelld3v:
Not allowing unused variables is good, but it shouldn't be that hard!

jsfan:
In theory, a warning would be enough. In practice, there are a lot of coders out there who ignore warnings altogether...

PR #573, opened 2020-11-10
Problems with proxy

shelld3v:
Description
---------------

Bug fix with proxy

### DIFF
--- a/lib/core/argument_parser.py
+++ b/lib/core/argument_parser.py
@@ -105,6 +105,8 @@ class ArgumentParser(object):
 
             self.proxylist = open(options.proxyList).read().splitlines()
 
+            options.requestByHostname = True
+
         elif options.proxy:
             if (
                 options.proxy.startswith("http://") or options.proxy.startswith("https://") 
@@ -113,6 +115,8 @@ class ArgumentParser(object):
                 self.proxy = options.proxy
             else:
                 self.proxy = "http://{0}".format(options.proxy)
+
+            options.requestByHostname = True
 
         else:
             self.proxy = None


maurosoria:
I don't think request by name should be mandatory (the same reason why --ip is still useful) when using proxies.

Sometimes you want to use a proxy but setting directly the hostname.

shelld3v:
Have you ever tries to use the `--proxy` feature? I used Burp as a proxy and everything I saw in my dashboard is IP addresses. Use proxies, YOU CAN't SET THE HOST HEADER, and this will MAKE A LOT OF PROBLEMS (In most servers, if you don't set the host header to the correct hostname, you will unable to hit the website). I will say this is A SUPER IMPORTANT BUG FIX and don't know why it was there for years!

shelld3v:
What a normal code do:

```python
requests.get('http://google.com', proxies={'http':'127.0.0.1:8080'})
```

But what you do:

```python
requests.get('http://172.217.161.174', proxies={'http':'127.0.0.1:8080'})
```

And this is not what the user expects.

maurosoria:
Screen Shot 2020-11-11 at 14 42 35

I tested the feature, and it works as expected. As you see Burp you show the IP address in the table, but the HOST header is correct.

I don't know if you meant something else. Can you clarify?

shelld3v:
Oh, I didn't check the host header, sorry for that. But I believe it is really really important to set `requestByHostname`, as in a normal use case

shelld3v:
https://github.com/maurosoria/dirsearch/pull/556#issuecomment-723757637

PR #576, opened 2020-11-10
Bug fix

### LINKED ISSUE #575: [bug] UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb7 in position 6: invalid start byte
```
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/root/dirsearch/lib/core/Fuzzer.py", line 148, in thread_proc
    callback(result)
  File "/root/dirsearch/lib/controller/Controller.py", line 342, in matchCallback
    if excludeText in path.response.body.decode():
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb7 in position 6: invalid start byte
```

Continue to display what has been found

What it should be instead?

ubuntu 20, python3.8

shelld3v:
Description
---------------

Fix #575

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -491,13 +491,13 @@ class Controller(object):
             ):
 
                 for excludeText in self.excludeTexts:
-                    if excludeText in path.response.body.decode():
+                    if excludeText in path.response.body.decode('iso8859-1'):
                         del path
                         return
 
                 for excludeRegexp in self.excludeRegexps:
                     if (
-                        re.search(excludeRegexp, path.response.body.decode())
+                        re.search(excludeRegexp, path.response.body.decode('iso8859-1'))
                         is not None
                     ):
                         del path


maurosoria:
This is not a proper fix for this. You should first detect the enconding of the body (I guess requests already do this, but you can use chardet directly) and the decode with the proper format.

shelld3v:
The bug that I searched on Google was because the text contains a character that's not utf-8:

```python
>>> text = b'\xff\x32'
>>> text.decode()
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
>>>
```

Use iso8859-1 decode, you will able to decode any character even is it readable or not! I don't know is this the best fix for it or not but I have searched for 15 minutes and saw this is the only fix (there is a solution that by encoding it with utf-8 first then decode but we can't do it yet because this is bytes). I think you can just merge it since it can RESOLVE the problem for sure and i didn't any further problem that it can't work for sure.

shelld3v:
Why you should find the decoding method for the encoding method when there is a decoding method that can decode any encoding method =)

maurosoria:
Hello @shelld3v ,

By default, Python3+ uses utf8 as a default encoding. For that reason, it's recommended to transform everything to utf8 (to avoid problems).

check this stackoverlfow:
https://stackoverflow.com/questions/58568285/how-does-requests-determine-the-encoding-of-a-reponse


As you see requests attempts to decode using the format provided by the Content-Type header. However, from my experiencie, I can tell you that the enconding in Content-Type header and the body of the response aren't always the same.

If you see other projects (like sqlmap) this have been a problem forever, and the best thing to do is:
* Check if the response is already UTF8
* if not, detect the encoding (with chardet.detect)
* Encode the string properly to UTF8 (with errors='replace' argument)

shelld3v:
No, there won't be any problem here. Whenever there is one, I will try to fix it. But until now, I only saw this is the best solution.

About your information, there is no way for us to identify the encoding type in this situation, because this is bytes: `b'\xff\x61\x32'`. And the problem is when we decode it to text, python3 can't transfer `'\xff'` (`0xff`) to UTF-8 text (UTF-8 is set by default in Python3 as what you said), because in UTF-8, there is no "ÿ" character. So what I do is transfer it to iso8859-1 text, which allows every character (with any byte).

shelld3v:
Visit this: https://www.urldecoder.org/ and try to decode `%ff` with UTF-8, you will receive the error. But change the source character to ISO, you will able to see it.

maurosoria:
let's see what happens

PR #583, opened 2020-11-13
Add requirements.txt

wdahlenburg:
Description
---------------

Adding a requirements.txt file to include dependencies. 

A fresh download of dirsearch can ensure it has the proper packages by running the following:
```pip3 install -r requirements.txt```

The dependencies were meant to match the code in https://github.com/maurosoria/dirsearch/blob/master/thirdparty/requests/__init__.py

Feel free to bump any of the versions or add dependencies if I missed some.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+certifi>=2020.11.8
+chardet<3.1.0
+urllib3<=1.25
+cryptography>1.3.4


shelld3v:
Hi, I think that `urllib3` has already been installed in Python 3 as well as `chardet`, `cryptography` and `certifi`

shelld3v:
I am thinking about add `PySocks` into this file because I tried to add it into third-party but it does not work, and `PySocks` is important to can use SOCKS proxy (https://github.com/maurosoria/dirsearch/blob/master/CHANGELOG.md). But I am still trying to fix that problem and haven't wanted to create requirements.txt yet since it's really hard for Windows user to import python pips

wdahlenburg:
Sure. The versions may not be correct though for any of the packages, which could lead to inconsistencies.

The requirements.txt file is useful for anyone using python's virtual environments.

PR #599, opened 2020-11-16
[End-of-life] Remove no-dot-extensions option because it's useless

shelld3v:
Description
---------------

Since people can now use suffixes and prefixes for this, I decided to end NDE because I think no one is actually using this option and dirsearch is going to be implemented a huge number of new arguments in the near future.

So, let's say "good-bye" to no-dot-extensions.


shelld3v:
Good bye 😢

maurosoria:
😭

shelld3v:
🤣

PR #610, opened 2020-11-19
Clean code

shelld3v:
Description
---------------

Clean code =)


maurosoria:
I love you so much @shelld3v ❤️ 

I think 'not n' is not as clear as 'n == 0' in this particular case.

shelld3v:
"not n" means "n == 0 or n == None or not len(n)". I don't know if it can make any problem here?

shelld3v:
If you don't like this, you can close this! Since it's not important at all, I just had so much free time :P

PR #621, opened 2020-11-21
Smarter recursion

shelld3v:
Description
---------------

Something like this shouldn't be added into `directories`:

```
/+CSCOT+/oem-customization?app=AnyConnect&platform=..&resource-type=..&name=%2bCSCOE%2b/portal_inc.lua&type=oem
```

Otherwise, dirsearch will scan this:

```
/+CSCOT+/oem-customization?app=AnyConnect&platform=..&resource-type=..&name=%2bCSCOE%2b/portal_inc.lua&type=oem/FUZZING
```

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -523,10 +523,7 @@ class Controller(object):
                         if subdir == path.path + "/":
                             pathIsInScanSubdirs = True
 
-                if not pathIsInScanSubdirs:
-                    if not self.recursive:
-                        pass
-
+                if not self.recursive and not pathIsInScanSubdirs and "?" not in path.path:
                     elif path.response.redirect:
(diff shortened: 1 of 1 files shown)


stefanocoding:
This commit breaks https://github.com/maurosoria/dirsearch/blob/master/lib/controller/controller.py. Line  531 and 532.

shelld3v:
Hi, #622 will fix this. If you want to use dirsearch right now, you can try: `git clone https://github.com/TheBinitGhimire/dirsearch.git`

stefanocoding:
Hi @shelld3v.  Thanks for the reply! 
I commented just in case nobody noticed it :smile:. I was able to fix it locally temporarily until the next release.

PR #631, opened 2020-11-24
Clean code + reduce startup delay

shelld3v:
Description
---------------

This will allow us to skip the second test request if the first response gives us 404!

### DIFF
--- a/lib/core/scanner.py
+++ b/lib/core/scanner.py
@@ -108,9 +108,6 @@ class Scanner(object):
             if re.match(self.redirectRegExp, response.redirect) is None:
                 return True
 
-        elif self.redirectRegExp or response.redirect:
-            return True
-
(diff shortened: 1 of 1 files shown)


shelld3v:
Hi, any problem with this? I can adjust if you want!!

maurosoria:
Hello @shelld3v, I'm back. Hope you missed me 👍 

Can you add the same logic for the redirect regex, but for Set-Cookies too ? I always wondered what might happend.

shelld3v:
>Can you add the same logic for the redirect regex, but for Set-Cookies too ? I always wondered what might happend.

Added into TODO 👍 

>Hello @shelld3v, I'm back. Hope you missed me 👍

100 percent

shelld3v:
Think again, I may don't check the Set-Cookies header, because in some web apps, the Set-Cookies may be changed after each request!

PR #504, opened 2020-10-18
[Help needed] 429 Detector

shelld3v:
Description
---------------

Hi everyone, I am trying to create a 429 Detector, to finish one of the ideas I have already thought about for this project. But currently, I am having a problem with this:

```
$ python3 dirsearch.py -E -u localhost:4444

  _|. _ _  _  _  _ _|_    v0.4.0
 (_||| _) (/_(_|| (_| )

Extensions: php, asp, aspx, jsp, html, htm, js | HTTP method: GET | Threads: 20 | Wordlist size: 9990

Error Log: /tmp/dirsearch/logs/errors-20-10-18_13-28-37.log

Target: http://localhost:4444/

Output File: /tmp/dirsearch/reports/localhost/_20-10-18_13-28-37.txt

[13:28:37] Starting:
429 Too Many Requests detected: Server is blocking requests
11.99% (1198/9990) - Last request to: 52e
[stop response from here]
```

Correct:

```
$ python3 dirsearch.py -E -u localhost:4444

  _|. _ _  _  _  _ _|_    v0.4.0
 (_||| _) (/_(_|| (_| )

Extensions: php, asp, aspx, jsp, html, htm, js | HTTP method: GET | Threads: 20 | Wordlist size: 9990

Error Log: /tmp/dirsearch/logs/errors-20-10-18_13-28-37.log

Target: http://localhost:4444/

Output File: /tmp/dirsearch/reports/localhost/_20-10-18_13-28-37.txt

[13:28:37] Starting:
429 Too Many Requests detected: Server is blocking requests
[c]ontinue / [e]xit: e

Canceled by the user
```
 Hope if anyone out there can help to fix this issue. For more information, I think the problem comes from here:

```
    def processPaths(self):
        while True:
            try:
                while not self.fuzzer.wait(0.3):
                    continue
                break

            except (KeyboardInterrupt, SystemExit):
                self.handleInterrupt()
```

Somehow we need to stop this loop, is that correct???

PR #633, opened 2020-11-24
Handle 429 Status Codes

wdahlenburg:
Description
---------------

429 Status Codes indicate the rate limit was exceeded. #504 was created as the initial solution of handling this. I added a few helper functions and resolved the deadlock. 

Moving the PR from (https://github.com/shelld3v/dirsearch/pull/2) to here. 

The jist of this is to allow the threads to be paused once a 429 is received. The user can decide to continue, exit, or skip the host. If the user continues, execution resumes and they are acknowledging that 429 status codes have been received, which will not be returned in the output after that point.


shelld3v:
We will need to check this before merging!

shelld3v:
And @wdahlenburg, you need to add this feature information to `CHANGELOG.md`

shelld3v:
Or we can add this after releasing dirsearch v0.4.1 if you want @maurosoria

maurosoria:
Thanks to @wdahlenburg for helping @shelld3v ! 

Love the fact that two people worked together for this particular feature =)

PR #638, opened 2020-11-25
New arguments group: Request Settings

shelld3v:
Description
---------------

Reduce incorrect-thinkings from dirsearch users who don't read the doc


shelld3v:
I just created a new arguments group: Request Settings. I do this because the General is too big and will be even bigger in the future. I will create the `Technique` group if it's going to be large again!

PR #651, opened 2020-12-06
Print only used extensions

shelld3v:
Description
---------------

This PR will help the users have a better understanding of what is happening with their wordlists in the backend. If the user wordlists come from SecLists or something, that doesn't have the classic %EXT% keyword, dirsearch will let the user know that no extension is used. And if there is %EXT%, or the users use --force-extensions, dirsearch will print the extensions.

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -246,7 +246,7 @@ class Controller(object):
     def printConfig(self):
 
         self.output.config(
-            ', '.join(self.arguments.extensions),
+            ', '.join(self.dictionary._extensions),
             ', '.join(self.arguments.prefixes),
             ', '.join(self.arguments.suffixes),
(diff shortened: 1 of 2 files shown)


maurosoria:
Hello @shelld3v ,

I think this additional step is a good idea, but this current implementation wouldn't be performant. A good idea is try to look into the wordlist, but with a regex or with some other faster mechanism.

Regards,
Mauro

shelld3v:
Hey, I don't understand! Is this too slow? Hmm, probably not, since reading some files and check if `%EXT%` in the content is just a simple python process (takes < 0.1 sec)

shelld3v:
Any idea?

shelld3v:
@maurosoria

shelld3v:
Using regex here will lead to a bad performance

PR #652, opened 2020-12-06
Disable auto-save report by default

shelld3v:
Description
---------------

Disable auto-save report by default


maurosoria:
Mmmm I don't get why we should change this behaviour. Is there any reason?

shelld3v:
Idk, but the output looks a little bit messy with me

PR #698, opened 2020-12-29
Add git hooks and git objects pack info.

catmandx:
Added all the git hooks file, git/objects/info/packs (packfiles names) and git/info/refs.

Description
---------------

I was doing a lab and find that this tool doesn't list some important files like the file that contains the names of available packs and git hooks files.

---------------


maurosoria:
Thanks @catmandx, I appreciate =D

PR #701, opened 2020-12-30
New feature: URLs from stdin

shelld3v:
Description
---------------

URLs from stdin. Example command:

```bash
$ amass enum -d yahoo.com -o yahoo.txt && cat yahoo.txt | httprobe -prefer-https | python3 dirsearch.py --stdin
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 Currently working...
 ---------
-- Fixed important bugs from v0.4.1
+- Fixed critical bugs from v0.4.1 (IMPORTANT)
 - Exclude responses by redirects
+- URLs from STDIN
 
 v0.4.1 (2020.12.8)
 ---------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,8 +100,21 @@ v0.4.1 (2020.12.8)
 
 ---------
 
-- 0.2.6 - 2014.9.12 Fixed bug when dictionary size is greater than threads count. Fixed URL encoding bug (issue2).
-- 0.2.5 - 2014.9.2 Shows Content-Length in output and reports, added default.conf file (for setting defaults) and report auto save feature added.
+0.2.6 - 2014.9.12
+---------
+- Fixed bug when dictionary size is greater than threads count
+- Fixed URL encoding bug
+
+---------
+
+0.2.5 - 2014.9.2
+---------
+- Shows Content-Length in output and reports
+- Added default.conf file (for setting defaults)
+- Report auto save feature added.
+
+---------
+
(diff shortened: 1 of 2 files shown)


maurosoria:
Hello @shelld3v, how you doing?

First, I recommend to use splitlines instead of split("\n"), or directly use FileUtils getLines with stdin.

I have another question, did you do proper testing with large files?

shelld3v:
Hi, `sys.stdin.read()` will give you a string object, there isn't the attribute `readlines()`.

>I have another question, did you do proper testing with large files

Yes

maurosoria:
It's not readlines, is splitlines

Check the main function https://github.com/maurosoria/bugbounty-tools/blob/master/resolvedomains/resolvedomains.py

You can do it with FileUtils and sys.stdin

maurosoria:
Another thing, most of tools use the '-' character to refer to the stdin. For example:
```
cat urls.txt | dirsearch -l - 
or
cat wordlist.txt | dirsearch -u target.com -w -
```

The stdin argument seems fine, but if you want to make it more fancy, the - syntax it's more familiar.

Just a suggestion, all solutions seem find to me.

shelld3v:
> Another thing, most of tools use the '-' character to refer to the stdin. For example:
> 
> ```
> cat urls.txt | dirsearch -l - 
> or
> cat wordlist.txt | dirsearch -u target.com -w -
> ```
> 
> The stdin argument seems fine, but if you want to make it more fancy, the - syntax it's more familiar.
> 
> Just a suggestion, all solutions seem find to me.

Let's see



> Another thing, most of tools use the '-' character to refer to the stdin. For example:
> 
> ```
> cat urls.txt | dirsearch -l - 
> or
> cat wordlist.txt | dirsearch -u target.com -w -
> ```
> 
> The stdin argument seems fine, but if you want to make it more fancy, the - syntax it's more familiar.
> 
> Just a suggestion, all solutions seem find to me.

Nice, I will check it!

shelld3v:
After thinking again, I decided not to make the `-` syntax as a signal to read URLs from stdin. Because it will be hard to notify people about that syntax (not everyone likes reading the README). And actually, not most of the tools use the '-' character to refer to the stdin, it was just an idea from GoBuster, then FFUF copied it (FFUF was inspired by GoBuster and wfuzz), feroxbuster is an example of not using that syntax.

Separate the `--stdin` flag will make things work easier!

maurosoria:
Hello @shelld3v !

I merged the PR. The - syntax is not a gobuster idea (didn't know gobuster works that way), it's a convention for plenty of unix tools. 

https://unix.stackexchange.com/questions/16357/usage-of-dash-in-place-of-a-filename

shelld3v:
Well, I will soon look again at this!

PR #702, opened 2020-12-30
Update README.md and more

shelld3v:
Description
---------------

From the closed PR :)

What do you think about the logo and the *Hacking is not a crime* message? I didn't think about adding that but got a too huge empty space in the right so ... this message can be an inspiration :new_moon_with_face:

### DIFF
--- a/.github/ISSUE_TEMPLATE/ask_question.md
+++ b/.github/ISSUE_TEMPLATE/ask_question.md
@@ -1,6 +1,6 @@
 ---
 name: Ask Question
-about: Ask a question about Dirsearch
+about: Ask a question about dirsearch
 labels: question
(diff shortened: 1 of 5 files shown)


maurosoria:
What hacking does mean to you? For me, is the art of understanding how things work and creating new things (like an exploit, a tool or even an electronic device) from what we learnt.

Indeed, it's not a crime, it's art

shelld3v:
Yeah, the same, but with me it's not creating new things, just breaking computer stuff in a super logical and creative way!

maurosoria:
Hello shelld3v

I merged the Readme, but the logo is unreadable in Github's dark mode (probably a purple just like the banner?)

PR #707, opened 2021-01-03
Don't add / when force extensions

shelld3v:
Description
---------------

From a user, he said that the wordlist is too big when using the `-f` flag: 200k -> 600k. I think that we don't need to add the slash when forcing extensions, if the users want to brute-force with the slash, they can use `--suffixes`. 

In a normal case (from my experience), if you hit `folder`, then it will redirect you to `folder/` and we will know `folder/` exists, but hitting `folder/` most of the times won't give you an interesting response.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ index.asp
 index.aspx
 ```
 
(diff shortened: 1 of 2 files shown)


maurosoria:
Hello @shelld3v ,

Actually, is very normal getting a hit with / at the end, and a 404 without it. Actually, forced extensiones didn't append the slash at the beginning, I modified to do so.

shelld3v:
@maurosoria In my whole experience, I only saw 2 ways it could happen:
- Hitting `admin` -> `admin/`
- Hitting `admin` or `admin/` will give you nothing, except for `admin/path` (Yahoo is that you?)

PR #708, opened 2021-01-04
Update the 429 message

shelld3v:
Description
---------------

Update the 429 message!

Btw, I am seeing that the dirsearch's folders, files, and codes are the same as sqlmap at some points!!! Is that accurate? ;)

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -648,7 +648,7 @@ class Controller(object):
             try:
                 while not self.fuzzer.wait(0.25):
                     if not self.ignore429 and self.got429:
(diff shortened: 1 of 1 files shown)


maurosoria:
Hello @shelld3v !


> Btw, I am seeing that the dirsearch's folders, files, and codes are the same as sqlmap at some points!!! Is that accurate? ;)

The only part I used from sqlmap is in the thirdparty folder. About folders and files, I don't know, I just simply used MVC and started to change the file structure according to what I considered proper. I've studied sqlmap code in the past, but actually it's so large and complex that is a nightmare to follow.

PR #725, opened 2021-01-13
Bug fix with headers parsing

shelld3v:
Description
---------------

The bug will be caused by headers like this: `Header: Var::Value`

Here is the fix


shelld3v:
I just finished all the exams with not bad marks! So I am going to super focus on hacking (bug bounty) and updating dirsearch 👍

shelld3v:
Please release dirsearch v0.4.2 =)

PR #749, opened 2021-01-31
Several updates

shelld3v:
Description
---------------

A lot of updates and bug fixes


maurosoria:
Hello @shelld3v , how you doing?

Can you explain and describe the PR? There are some changes that I don' get them..

shelld3v:
So, dirsearch doesn't check for wildcard results when I use suffixes/prefixes. (Example: `--suffixes ~ -u example.com` will lead to a 403 flood), I fixed it. Also, the `--calibration` flag doesn't work as expected, so made it A LOT better, renamed it to `--exclude-content` as well. Moreover, rather than returning only 1 tester (`fuzzer.py`), it will now return several ones, for better checking, example: `.index.php` will return `scanner['php']`, `scanner['dotfiles']` and `defaultScanner`. 

I also edited the `generateRedirectRegExp` function, rather than use `SequenceMatcher` from `difflib`, I make a really good diff comparator that will generate a better regex (compare with the old regex, it seems to reduce more false positives and false negatives, I even did a test)

shelld3v:
I am trying to add more comments to the code so new contributors won't face too much problems, but it's so hard for me ;(

maurosoria:
I==

> So, dirsearch doesn't check for wildcard results when I use suffixes/prefixes. (Example: `--suffixes ~ -u example.com` will lead to a 403 flood), I fixed it. Also, the `--calibration` flag doesn't work as expected, so made it A LOT better, renamed it to `--exclude-content` as well. Moreover, rather than returning only 1 tester (`fuzzer.py`), it will now return several ones, for better checking, example: `.index.php` will return `scanner['php']`, `scanner['dotfiles']` and `defaultScanner`.
> 

I thinks this is a great solution. Actually I thpught about this in the past and  the main problem of using 1 scanner, but actually thought a way of generating different scanners based on the responses of previous matched paths (to avoid repeated resuts).

This is something similar and I like it, though I couldn't test it yet.

> I also edited the `generateRedirectRegExp` function, rather than use `SequenceMatcher` from `difflib`, I make a really good diff comparator that will generate a better regex (compare with the old regex, it seems to reduce more false positives and false negatives, I even did a test)
>
About this change, I wonder why you want to extract SequenceMatcher from the equation. It's a flexible way of generating regex based on the variations of the responses. 
Another thing, why are you trying to escape manually the characters? It seems to me you're trying to rewrite re.escape and I don't get it.

shelld3v:
>About this change, I wonder why you want to extract SequenceMatcher from the equation. It's a flexible way of generating regex based on the variations of the responses.

I tested the old one and found that it will filter more than it should be, but still have false positives. So with my new one, it's a lot better.

>Another thing, why are you trying to escape manually the characters? It seems to me you're trying to rewrite re.escape and I don't get it.

lol I didn't know that

shelld3v:
Done?

shelld3v:
>I thinks this is a great solution. Actually I thpught about this in the past and the main problem of using 1 scanner, but actually thought a way of generating different scanners based on the responses of previous matched paths (to avoid repeated resuts).

I am going to make a duplicates checker

shelld3v:
> > I thinks this is a great solution. Actually I thpught about this in the past and the main problem of using 1 scanner, but actually thought a way of generating different scanners based on the responses of previous matched paths (to avoid repeated resuts).
> 
> I am going to make a duplicates checker, but it's ok to merge now

Done

maurosoria:
Hello @shelld3v ,

I still don't get it why you want to remove the original regex generator for redirects. I know it looks messy, but it worked. 

And I don't like the thing about replacing with "DS_PATH" (what might happend if someone put that in their wordlist).

Can you explain to me? Provide me with test scenarios?


Regards, 
Mauro

> > > I thinks this is a great solution. Actually I thpught about this in the past and the main problem of using 1 scanner, but actually thought a way of generating different scanners based on the responses of previous matched paths (to avoid repeated resuts).
> > 
> > 
> > I am going to make a duplicates checker, but it's ok to merge now
> 
> Done

About this, I guess you misunderstand me. What I meant is the following behaviour:
* Dirsearch executes with some extensions, for example: -e php
* There are different extensions in the wordlist, for example: .old .jsp .asp
* There are different suffixes or troublematic words: for example .ht.* family and paths with the word 'admin'
* There are no scanners preconfigured for these.
* But Fuzzer Class has the responses in self.matches (response included).
* Fuzzer could check if there is an scanner extension for the self.matches (or a suffix). If there are two matched with that extension/suffix, Fuzzer could try to create an additional scanner for that type of paths in runtime.

Note: this shouldn't be a default option, but it could reduce a lot the false positives.

shelld3v:
>I still don't get it why you want to remove the original regex generator for redirects. I know it looks messy, but it worked.

No, it's not working at all. I tested with `http://ffuf.io.fi`, which always redirects you to `https://ffuf.io.fi`. And in that test, `http://ffuf.io.fi/php` was valid, because the regexp was `https://ffuf.io.fi/.*.php`! That's a false positive, right? What happens if the `http://ffuf.io.fi/internal.php` redirects to `http://ffuf.io.fi/privateKeyword.php`? Filter it? That's false negatives!! So my update is really important

>About this, I guess you misunderstood me. What I meant is the following behaviour:

Dirsearch executes with some extensions, for example: -e php
There are different extensions in the wordlist, for example: .old .jsp .asp .bak .swp
There are different suffixes or troublematic words: for example .ht.* family and paths with the word 'admin'
There are no scanners preconfigured for these (only the one provided by -e and the default ones).
But Fuzzer Class has the responses in self.matches (response included).
Fuzzer could check if there is an scanner extension for the self.matches (or a suffix). If there are two matching with that extension/suffix, Fuzzer could try to create an additional scanner for that type of paths in runtime.

That's nice, I will try to implement it in the future, But remember that, this will generate a long delay

shelld3v:
>And I don't like the thing about replacing with "DS_PATH" (what might happend if someone put that in their wordlist?).

Hey, I updated the code a little bit, for more accuracy and replaced `DS_PATH` with `DIRSEARCH_PATH`, I don't think there will be someone put this into their wordlist. I can't find anything else I can do, cause the replacement is the only way to reduce the false negatives (worse than false positives). I didn't want to update the Regex generator, but it did miss my stuff, so this is a must-do thing!

shelld3v:
Even if there is really someone put this into their wordlist, probably that is a troll by someone else, then a single false positive isn't a big deal!

shelld3v:
Any update for my PR? :)

maurosoria:
Hello @shelld3v , how you doing?

Please, divide the PR (the DS_PATH and Location Regex) from the other patches. That'd be easier for testing and reverting.

shelld3v:
> Hello @shelld3v , how you doing?
> 
> Please, divide the PR (the DS_PATH and Location Regex) from the other patches. That'd be easier for testing and reverting.

Hi @maurosoria, I'm good, what about u?

Sorry for combining it, I afraid that separating them would make conflicts for others when one is merged

shelld3v:
Btw, I removed DS_PATh with a random number as the sign

PR #761, opened 2021-02-20
Include cmdline args in reports

wdahlenburg:
Description
---------------

This PR includes the cmdline arguments for the json, markdown, plain, and xml reports. This was attempted to be done in a format similar to nmap. This is helpful metadata to include in reports because users can better understand output based on their flags. It's difficult to always remember what filters you used, so automatically including it in reports maintains that info.

The CSV and simple formats were left out. A metadata line could be added to both formats, but I felt like it would be extraneous info/parsing for users wanting these formats. 

The JSON and XML formats were restructured a bit. I'll open up a separate issue, but I think there are some general improvements that could be made to the structure. Feel free to modify or suggest a different structure for this PR.

### DIFF
--- a/lib/reports/json_report.py
+++ b/lib/reports/json_report.py
@@ -18,6 +18,7 @@
 
 import json
 import time
+import sys
 
 from lib.reports import *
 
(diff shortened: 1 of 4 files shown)


shelld3v (approved):

shelld3v:
Hi, from what I know, Mauro is having some issues and can't merge PRs right now. Please be patient

PR #785, opened 2021-04-12
Refactored the wildcard redirect regex generator

shelld3v:
Description
---------------

From #749


shelld3v:
Hi Mauro, if you released v0.4.1-alpha2, then should you add all new features to the release summary?

PR #810, opened 2021-04-25
WIP: detection templates

shelld3v:
Description
---------------

Work in progress... (info will be updated soon)


maurosoria:
Hello @shelld3v,

I guess the correct term of what you are trying to do is "signatures" and not "detection templates".

shelld3v:
> guess the correct term of what you are trying to do is "signatures" and not "detection templates".

Should there be a poll? 🙄

shelld3v:
This name was taken from nuclei (`nuclei-templates`)

shelld3v:
What about **signature-templates**❓

shelld3v:
Now we just need for default templates and a good documentation

maurosoria:
Hello @shelld3v,

First, I guess "signatures" is a one word definition. You can then define different "templates" for "signatures". But the correct term is "signature".

Then, you are breaking the MVC pattern putting logic in the controller. What I think would be the best, is the fuzzer class , just append additional callbacks in Fuzzer.matchCallbacks. You should consider defining a "Signature" class, and embed all the related logic in there. Then, it would be just matter of adding that callback to the fuzzer (through the Controller, for example).

signatures/directory_listing
```diff
@@ -0,0 +1,13 @@
...
+body: 
+  regex: >
+    ^(.*)<html>
+     <head>
+      <title>Index of /(.*)</title>
+     </head>
+     <body>
+    <h1>Index of /(.*)</h1>
+      <table>(.*)$
+
```
wdahlenburg, line 13: What about something like this?

```yaml
name: Directory Listing

request:
  method: GET

response:
  status: 200
  body: 
    regex: >
      ^(.*)<html>
       <head>
        <title>Index of /(.*)</title>
       </head>
       <body>
      <h1>Index of /(.*)</h1>
        <table>(.*)$
```

It would be good to break out the request and response portions. This would allow the author of a signature to specify when it should be applied. There are a handful of characteristics that would be important.

* http-verb
* url contains or equals
* header exists
* request body contains
* response status code
* response header exists
* body contains
* body length is greater or less than XYZ

There will be less responses to check as more criteria is defined in the request section.

Muniyang:
Next time no message for me ok

signatures/directory_listing
shelld3v: Thanks, I will remember that. This code is old, I haven't worked on this for a while and it looks bad. I will refactor the whole thing when I'm back to work

PR #826, opened 2021-05-07
Release dirsearch v0.4.2

shelld3v:
Description
---------------

Release dirsearch v0.4.2 🎉


shelld3v:
>Is mandatory to check the new dependencies (HttpNtlmAuth make use of some not builtin libraries). Best way is creating a virtualenv or a fresh box and test if everything works without the requirements.

I tested it, and it worked

>Testing HTML report and new report system (specially about mako templates, I don't know the security concerns or best practices)

I will, hope no SSTI will be found

>I guess we should wait for v0.4.3 for signatures and templates ...

What about v0.4.5 or v0.5.0? Want to raise it faster for v1.0.0

shelld3v:
What about now?

maurosoria:
@shelld3v !

I guess we should stop adding functionality. I'm concerned about the text "decoding" to Iso-8859-1. I speculate issues with that.

If do you agree, stop adding functionality until the 1st of June. If everything goes right, we release v.0.4.2 that day. Do you agree?

shelld3v:
> @shelld3v !
> 
> I guess we should stop adding functionality. I'm concerned about the text "decoding" to Iso-8859-1. I speculate issues with that.
> 
> If do you agree, stop adding functionality until the 1st of June. If everything goes right, we release v.0.4.2 that day. Do you agree?

No problem

shelld3v:
Ping

shelld3v:
Bruh where is the release?

PR #850, opened 2021-05-21
Several updates

### LINKED ISSUE #753: New Suggestion
Hello Team , 
So I have a suggestion for the dirsearch , so from my experience , whenever I run run dirsearch on a target and it throws a file greater than 10-15 mbs , the bruteforced directories doesn't work , so can you add a feature to like discard the target if the file exceeds 10 mb ? That would be a really cool feature in my opinion , what do you think ?

shelld3v:
Description
---------------

- Move get_blacklists to Dictionary
- Stream response instead of download into the memory
- Only force extensions on paths don't have extension (this was declined in the past, but want to talk about it again). Example:

```
admin
index.html
```

`dirsearch -e jsp -f -u http://target`:

```
admin
admin.jsp
admin/
index.html # Without my update, it will be index.html.jsp, index.html/ and index.html
```

Fix #753


shelld3v:
What do you think @maurosoria?

maurosoria:
Hello @shelld3v !!!

Why are you decoding the response body in the Path ? requests automatically decode the content using chardet.

I guess that encapsulating the Blacklist in the Model could be good, but receiving ```script_path``` as a constructor argument is not okay. It should receive something as file_paths. 

Although the previous, I like the PR.

shelld3v:
>Why are you decoding the response body in the Path ? requests automatically decode the content using chardet.

We need to decode it to 'iso8859-1' so that regex can work probably (there is a bug submitted for this by RandomRobbie)

PR #859, opened 2021-05-25
Add CVE-2021-27651 check

ricardojba:
Add check for Pega infinity CVE-2021-27651
REF: https://github.com/samwcyo/CVE-2021-27651-PoC/blob/main/RCE.md


shelld3v (approved):
Thank you

PR #860, opened 2021-05-25
Stream response instead of download into the memory

### LINKED ISSUE #753: New Suggestion

shelld3v:
Description
---------------

By default, `requests` downloads the whole response at a time to an object, but if the response is too huge, it will hit the memory painfully. The result of this is dirsearch gets stuck, so I decided to stream response, which means get response data bytes by bytes.

Fix #753

### DIFF
--- a/lib/connection/requester.py
+++ b/lib/connection/requester.py
@@ -191,15 +191,11 @@ class Requester(object):
                     proxies=proxies,
                     allow_redirects=self.redirect,
                     timeout=self.timeout,
+                    stream=True,
                     verify=False,
                 )
 
-                result = Response(
-                    response.status_code,
-                    response.reason,
-                    response.headers,
-                    response.content,
-                )
+                result = Response(response)
 
                 break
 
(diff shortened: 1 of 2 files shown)


maurosoria:
Hello @shelld3v 

Directly decoding a chunk can bring potential errors. There is not even a try catch to manage de error. Using chardet could also be good.
```
self.body += chunk.decode("iso8859-1")
```

If you really want to do this well, its also important to parse the Response Content-Length header (applying a limit to the buffered chunks is important too).

shelld3v:
> Directly decoding a chunk can bring potential errors. There is not even a try catch to manage de error. Using chardet could also be good.

This chunk is just a byte-type object

>If you really want to do this well, its also important to parse the Response Content-Length header (applying a limit to the buffered chunks is important too).

Sounds good, but why? I may consider this in order to can see the response length when performing HEAD requests

shelld3v:
>Sounds good, but why? I may consider this in order to can see the response length when performing HEAD requests

But we will need to merge #850 forst

shelld3v:
I'm going to merge this into #850

PR #873, opened 2021-05-28
F401/F403 Eliminate wildcard imports

jsfan:
Description
---------------

This PR eliminates all wildcard imports and enables checking for rules [F401](https://www.flake8rules.com/rules/F401.html) and [F403](https://www.flake8rules.com/rules/F403.html) on the CI.


shelld3v (approved):
Thanks

PR #883, opened 2021-06-06
Update ratio adjustment to improve the accuracy

shelld3v:
Description
---------------

Update ratio adjustment to improve the accuracy

### DIFF
--- a/lib/core/scanner.py
+++ b/lib/core/scanner.py
@@ -96,12 +96,12 @@ class Scanner(object):
         if first_path in first_response.body.decode():
             if len(first_response) < 200:
                 self.ratio -= 0.15 + 15 / len(first_response)
-            elif len(first_response) < 1000:
+            elif len(first_response) < 800:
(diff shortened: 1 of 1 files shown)


maurosoria:
Hello @shelld3v !

I'd like to point out that the detection of reflected content is a good idea, but in the way it was code in the PR you are giving for granted that the reflected URI is not encoded. In many cases, at least you have to check if it is URL encoded.

I suggest to use an OR conditional to check different encodings.

maurosoria:
Another escapings like JSON should be taking into account. I don't know if I miss another one important.

shelld3v:
>I'd like to point out that the detection of reflected content is a good idea, but in the way it was code in the PR you are giving for granted that the reflected URI is not encoded. In many cases, at least you have to check if it is URL encoded.

This has already been knowledged. The contents we are checking are from the test requests, and test requests will be: `stringonly`, `string.extension`, ... So there won't be any encoded path here in the test request contents (no slash, question mark, ...)

PR #912, opened 2021-07-02
Update default.conf

shelld3v:
Description
---------------

Update default.conf


maurosoria:
Js extension is a must.

shelld3v:
I just think that js files don't appear much outside a specific folder (`js/`, `static/`, `asset/`, `assets/`, ...)

PR #914, opened 2021-07-06
Enhanced Search + Bug Fix in Report Template

sh33lsh00k:
Description
---------------

**What will it do?**
For an advanced search, this pull request contains 2 new search fields:
- Exclude Status Code
- Exclude Content-Length

All search fields can be used simultaneously. Many a time we get a lot of undesired results from dirsearch when non-existent endpoint returns status code other than 404 e.g. 200 OK, 3xx, etc
To hide undesired results I have included 2 new search fields. User can exclude multiple status code and content length separated by space.

**Example Search:**
Shows result that
- Includes "text/html"
- Exclude 302 301 404 status code
- Exclude 0B Content Length

Also added Open URLs button to open the filtered results in new tab. (Chrome blocks opening of multiple new tabs - Click Always allow pop-ups)

I have also added some CSS for an improved look.

**Fix**
Added check for:
-  result.contentType as it is null when the content length is 0 bytes.
- Fixed typo result.statusColorClass to result.status_color_class

image


shelld3v:
😱😱😱Sick, that looks addicted. I am also thinking about sorting results by length/status/... What do you think? :)

shelld3v:
Btw, I think multiple status codes and content lengths should be separated by `,`

shelld3v (approved):

sh33lsh00k:
Hi @shelld3v 
Thanks for the suggestion, will update the report as soon as I find some time ‎😃

shelld3v:
Hi, I realized that your PR does have downsides, such as you removed our "colors for status codes" feature, the text color (black) is the same as the background color (dark), ...

shelld3v:
Actually the report in your screenshot is good, but I don't know why my chrome shows a different thing

image

shelld3v:
Oh sorry, it's not your fault, a commit by another contributor actually broke it

PR #920, opened 2021-07-08
Add Rack middleware exposure

ricardojba:
Rack middleware that provides access to rack.session environment in production
REF: https://nitter.unixfox.eu/jobertabma/status/1410756581239623681


shelld3v (approved):
Thanks 😉 👉👉

PR #942, opened 2021-07-30
fix decoding None err

XinRoom:
Description
---------------

Fix decoding None error   "utf-8" https://github.com/contact/report-content?content_url=https://github.com/maurosoria/dirsearch/issues/937#issuecomment-889905032


shelld3v (approved):
Thank you :heart:

PR #976, opened 2021-10-07
Addition to CVE-2021-41773

Xib3rR4dAr:
Added path for CVE-2021-41773 (Apache Path traversal vulnerability) which can be used if there is other alias set for for cgi-bin directory.
Reference: https://twitter.com/__mn1__/status/1445655933242134530


shelld3v:
Thank you!

PR #998, opened 2021-12-05
fix add_directory()

zrquan:
Description
---------------

What will it do?

In the original deep-recursive logic, if the path is a/b/c.html, only a/ will be added to the queue. Now it will add a/ and a/b/, and working fine with consecutive slashes.


shelld3v:
Have you tested this locally yet? I thought I tested that behaviour before and it worked well

shelld3v (approved):

zrquan:
> Have you tested this locally yet? I thought I tested that behaviour before and it worked well

Of course, I tested it locally

shelld3v:
Thanks

PR #1000, opened 2021-12-08
let dirsearch more intelligence,add domain combination dictionery for fuzz backup files

go-spider:
Description
---------------

let dirsearch more intelligence,add domain combination dictionaries for fuzz backup files
If target domain is admin.google.com,the dictionaries will add "admin.zip","admin.google.zip",etc for fuzzing test,It's very useful for real circumstances.

Requirements
---------------

- [ ] Add your name to `CONTRIBUTERS.md`

### DIFF
--- a/lib/core/fuzzer.py
+++ b/lib/core/fuzzer.py
@@ -18,6 +18,8 @@
 
 import threading
 import time
+import urllib.parse
+from itertools import combinations
 
 from lib.connection.request_exception import RequestException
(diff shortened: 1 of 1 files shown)


shelld3v:
There should be a separate tool for such things, not dirsearch. Or at least, we should add an option (command-line flag) for this, directly updating the wordlist without asking for user permission is unacceptable

PR #1039, opened 2022-01-29
Capability to save progress and bug fixes

shelld3v:
Description
---------------

Read the description

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,11 @@ jobs:
         pip install codespell flake8 -r requirements.txt
     - name: Test
       run: |
-        echo index.html > tmp
-        python3 dirsearch.py -u https://example.com --subdirs /,a/ -o tmp.txt --format json -r -w tmp -q -f -e html
+        echo index.%EXT% > tmp_wordlist.txt
+        echo testphp.vulnweb.com > tmp_targets.txt
+        python3 dirsearch.py -u https://example.com -o tmp_report.txt --format json --force-recursive -R 3 -w tmp_wordlist.txt
+        python3 dirsearch.py -l tmp_targets.txt --subdirs /,admin/ -X conf -q -L -f -i 200 --useragent abc --log tmp_log.txt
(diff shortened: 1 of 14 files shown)


maurosoria:
It would be really easy monitor the progress. This can be done by implementing some type of report (let's say, class "ProgressReport") so every time we process something with the wordlist, we get a callback whenever is processed (without modifying dirsearch, just using an extra report). 

Then, you can make a class called "ProgressState": this class should be attached to the new created report (previously mentioned ProgressReport) and it would consist just an index marking the processed items in the wordlist. Let's check the patator class for progress:

```
class ProgressState:  # this is a good example of how ProgressState 
  def __init__(self, dictionary):
    self.dictionary = dictionary
    self.found = []
    self.fail_errors = []
    self.seconds_passed = 0
    self.state_handlers = []
    self.processed_indexes = False * len(dictionary) # every processed callback should set to True the corresponding index of the wordlist

    def save_state(self):
        for state in  self.state_handlers:
            state.write_to_disk()
        # should serialize to disk "processed_indexes"

    def add_state_handler(self, handler):
        self.state_handlers.add(handler) # manage different kinds of handlers in disk
        

```

maurosoria:
So for a State class example:
```
class BaseState(object):
    def restore_from_disk(self, location):
        pass
        
    def write_to_disk(self):
        pass

...

class FileBasedState(BaseState):
    def write_to_disk(self):
        open('somefiletowrite', 'w')
        
...

class MysqlState(BaseState):
    def write_to_disk(self):
        mysql.insert("INSERT INTO ....")
```

shelld3v:
It's not just about how to save the progress but also how to continue the progress, without causing new problems/bugs. The classes you showed above only allow you to save some basic information of the scan, but not small things, and didn't show how to continue it, it will require a lot of work and effort to follow this path, but complex, insecure and unnecessary too. The solution I tried I think was the best as it doesn't require extra processes, doesn't monitor anything but just saves it and continues later, as it.

PR #1046, opened 2022-02-05
Release v0.4.2.2

shelld3v:


shelld3v:
Just a monthly release, no need to make a github release for it

its0x08 (approved):

maurosoria (changes requested):
global variables are note adviced

Use a BaseClass in the corresponding Report and ReportManager lib/ files.

maurosoria (changes requested):
```
            try:
                FileUtils.create_dir(self.report_path)

                if not FileUtils.can_write(self.report_path):
                    raise Exception
            except Exception:
                self.output.error(f"Couldn't create report folder at {self.report_path}")
                exit(1)
```

can_write why are you catching an Exception? is this a fix for some issue?

shelld3v:
> 

`Exception` gets raised whenever it fails to create directory (no write permission, can create, ...)

maurosoria (changes requested):
```
                except (InvalidURLException, RequestException, socket.gaierror) as e:
                    if e == socket.gaierror:
                        e.args[0] = e.args[1] = "Couldn't resolve DNS"
```

Using a low level library in a Controller is not a good practice.

Maybe a DNSError exception?

maurosoria:
> > 
> 
> `Exception` gets raised whenever it fails to create directory (no write permission, can create, ...)

Alright, let's catch it

shelld3v:
> global variables are note adviced
> 
> Use a BaseClass in the corresponding Report and ReportManager lib/ files.

But even we do that, we still have to `global` it right? For example the solution is:

```python
if not self.options["autosave_report"] and not self.options["output_file"]:
    global ReportManager, Report

    ReportManager = BaseReportManager
    Report = BaseReport
```

shelld3v:
@maurosoria what do you think about `from socket import gaierror`?

maurosoria:
> > global variables are note adviced
> > Use a BaseClass in the corresponding Report and ReportManager lib/ files.
> 
> But even we do that, we still have to `global` it right? For example the solution is:
> 
> ```python
> if not self.options["autosave_report"] and not self.options["output_file"]:
>     global ReportManager, Report
> 
>     ReportManager = BaseReportManager
>     Report = BaseReport
> ```

No, it is not necessary

maurosoria:
> gaierror

gaierror is a DNS Exception  (or RequestException). Would be better to wrap that exception as a RequestException and catch it from Requester.request

shelld3v:
> > > global variables are note adviced
> > > Use a BaseClass in the corresponding Report and ReportManager lib/ files.
> > 
> > 
> > But even we do that, we still have to `global` it right? For example the solution is:
> > ```python
> > if not self.options["autosave_report"] and not self.options["output_file"]:
> >     global ReportManager, Report
> > 
> >     ReportManager = BaseReportManager
> >     Report = BaseReport
> > ```
> 
> No, it is not necessary

I tried but it didn't work without `global`

maurosoria:
at lib/core/report.py, add in somewhere

```
class DummyReport(BaseReport):
    def save(self):
        pass

    def close(self):
        pass
```

then import that class from lib.core and assign the class

Another thing to point out is that Controller.setup_reports should be in charge of setting everything up for the reports (not the __init__).

Mostly the same for the reportmanager

maurosoria:
Hello @shelld3v ,

Using pickle for deserialization can be exploited to a rce. Please, look for a safer way of storing data. 

Probably you can check the following suggestion (and even support multiple formats):
* Google protobuf
* msgpack-python
* sqlite database (this is builtin in python, is db so is more flexible, but may be a little slower)

I am sorry I cannot approve this, but you know what would happen next day in exploit-db :s

maurosoria:
By the way, haven't checked all changes but my first impression is very good.

As a convention, when managing files:
* If the variable contains a path to a file/directory, use file_path (for example, output_file_path)
* If the variable contains a handler to the file, use file or fd (for example, output_file or output_file_fd)

Great work as always!

shelld3v:
@maurosoria The only way `pickle` can be exploited in this situation is that the user uses an untrusted session file from someone else, maybe we should warn people to aware of the session file they are going to run? Yes there are other ways around this but `pickle` produce the least data to save and best performance

shelld3v:
I can see many other projects like sqlmap also use `pickle`

maurosoria:
Hello @shelld3v ,

I still it isn't a good idea. I've checked the documentation but you are allowing thirdparty classes to be imported.

If they were only builtin types, maybe I would be convinced. Anyway, why you need to whitelist the thirdparty folder?

maurosoria:
About the PR, I'd like you to consider to split it because now is too big. Maybe you want to PR everything but the pickling stuff? That way I can now merge it and we can version it if you want.

shelld3v:
> Hello @shelld3v ,
> 
> I still it isn't a good idea. I've checked the documentation but you are allowing thirdparty classes to be imported.
> 
> If they were only builtin types, maybe I would be convinced. Anyway, why you need to whitelist the thirdparty folder?

Yeah I allow `thirdparty` class because we have `requests.Session()` object in `Requester`, I can't bring any security risk as no function in thirdparty allows code execution

shelld3v:
> About the PR, I'd like you to consider to split it because now is too big. Maybe you want to PR everything but the pickling stuff? That way I can now merge it and we can version it if you want.

I can wait

shelld3v:
Besides `requests.Session`, we also have `sqlmap.DynamicContentParser`, maybe we can make it more restricted by limit what modules are allowed (rn it's `thirdparty.requests` and `thirdparty.sqlmap`)

maurosoria:
Please, create a different PR because this one has 29 commits, it's too much, and if we want to trace a bug in the future, it could be more difficult to do.

I know this isn't very pleasant, but being organized is essential.

Make a PR for every change not related to the "magic import" or "save the state" feature.

Then PR the magic folder (I am not sure if nowadays it makes sense not entirely switching to requirements.txt instead of the thirdparty folder). Anyway, it is not a good idea to modify the code inside the package (remember what happened with requests the last time) like you are doing now with the compat folder.

About the sqlmap folder, I don't know if it still makes sense holding in the thirdpardy folder (the code by now should be really different, maybe with a comment pointing out to sqlmap reference would be enough).

If we eliminate the thirdparty folder, requirements.txt would make our life much easier. Moving the Sqlmap dependency to the core package (and not thirdparty folder).

If we do that, using another pickler like msgbox o google profobuf seems more appealing. Unserializing data stills dangerous, but also not using builtin types would be problematic (from the security perspective, but also from python portability issues).

If you stick to the idea of using pickle and the thirdparty folder, only whitelisted classes should be unpicked (CLASSES, not entire PACKAGES). I can tell you that whitelisting every class may be really problematic.


I'd like to know your opinion, and I hope you can understand why I can't merge this PR as it is now (please, it's great, splitting it won't be so hard).

shelld3v:
It's very hard to separate "pickle" related changes from other changes, because I changed the whole structure due to the new way of pickling/unpickling data. Also, I can't find any documentation about how to dump (save) objects using msgbox or google profobuf with python, I think I will stick with `pickle`, whitelisting all possible classes/builtins seems dirty but other options are worse for me

About `magic` import change, I have found a way to can make it work without fixing the import directly

shelld3v:
I think I'm tired of this, maybe it's time to remove the thirdparty. I'm curious, why did you make that folder in the first place?

shelld3v:
Allow all builtins except `exec`, `eval` and only `collections` (datatypes) or `requests` (no risk confirmed) modules, what do you think?

PR #1049, opened 2022-02-11
fix big zip file memory boom

lanyi1998:
Description
---------------

Avoid memory boom through Range header

---------------

- [ ] Add your name to `CONTRIBUTERS.md`

### DIFF
--- a/lib/core/fuzzer.py
+++ b/lib/core/fuzzer.py
@@ -179,6 +179,10 @@ class Fuzzer(object):
 
     def scan(self, path):
         wildcard = False
+        if path.find(".") != -1:
+            suffix = path.split(".")[-1]
+            if suffix in "zip,tar,rar,bz2,sql,gz":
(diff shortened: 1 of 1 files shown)


shelld3v:
I'm closing this PR as there are too many issues:
- Not only paths end with "zip,tar,rar,bz2,sql,gz" could return huge data, maybe a txt file could cause the same thing too
- You put the code in the wrong place
- Your solution is ugly and incomplete

I have a question, have you ever faced this issue in a real scan? Because I think I should have prevented this "memory boom" issue (it has been reported in the past) by making it download the body chunk by chunk. If the issue does exist then it's probably because of other parts that process the content, tell me if you did face this (using the latest version of dirsearch) so I can acknowledge

shelld3v:
I think the idea of using this header is good tho, feel free to add it to the **Tips** part in README (you can submit a PR for it)

PR #1071, opened 2022-03-19
fix: Jinja2 autoescape to avoid `XSS` and `XXE`

its0x08:
Description
---------------

What will it do?

It will turn auto escape for html reports to avoid possible XSS and use defusedxml instead of xml to avoid XEE

Requirements 
---------------

### DIFF
--- a/lib/reports/html_report.py
+++ b/lib/reports/html_report.py
@@ -32,7 +32,7 @@ class HTMLReport(FileBaseReport):
         file_loader = FileSystemLoader(
             os.path.dirname(os.path.realpath(__file__)) + "/templates/"
         )
-        env = Environment(loader=file_loader)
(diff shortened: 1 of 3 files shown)


shelld3v:
Please check the jobs

shelld3v:
I have tested this before and I'm sure there would be problems with XML report in your update, XXE isn't possible in that situation anyway only "XML bomb". Also, I have already noticed the `autoescape` option but I'm figuring out could we reduce any implemented XSS protection first before applying that option

shelld3v:
Thanks for your excitement anyway, you can join our slack to talk more about dirsearch if you want: https://join.slack.com/t/dirsearch/shared_invite/zt-15nzze7gz-EXBYx2_t9grxhFDvjxBuPw

its0x08:
Joined, thanks. 😁

PR #1155, opened 2022-07-30
Fix #1152

### LINKED ISSUE #1152: filter extensions for crawling
would love if you guys add a filter feature for crawling
ex : filtering js files

shelld3v:


maurosoria:
@shelld3v i have a conflict here! not sure if you want to keep the logging line

PR #1165, opened 2022-08-04
Add HTTP verbs to the dictionary

notdodo:
Description
---------------

What will it do?

- Add HTTP verbs to the dictionary
- ~Sort the dictionary~


shelld3v (changes requested):
Hi @notdodo, thanks for your effort but I would like to keep the old sorting method: special characters first then numbers and alphabetic characters, so basically: only add your new paths but keep the rest of the dictionary. An explanation for this is that many of our dictionary contributors (as well as me) use this method, we should keep the consistency of the dictionary.

notdodo:
Hi @shelld3v, thank you I didn't know about the ordering; while it's odd (for me :) ) this sorting it should be nice to have an _automated_ way to add new entries in the DB in the correct order and checking for duplicates.

Now it should all be good :)

shelld3v (approved):

PR #1172, opened 2022-08-19
Minor changes

shelld3v:


maurosoria:
Hello!

I see some incorrect default headers. Also, remember most browsers use headers with an uppercase at the beginning.

shelld3v:
@maurosoria dirsearch uses `CaseInsensitiveDict` for storing headers so it doesn't matter anyway. Plus HTTP headers are [officially] case-insensitive according to [RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616.html), I haven't seen any web framework or any server that behaves differently with capitalized headers

maurosoria:
Mmm please split the PR, the title says minor changes (but you modified 12 files)

shelld3v:
It's called minor changes because I only changed the design/structure, didn't impact user experience. Maybe I should rename it to "Major changes"?:)

PR #1227, opened 2022-10-30
Add more years to the wordlist

shelld3v:


maurosoria:
I have visions of %YEAR% keyword for the wordlist ....

PR #1232, opened 2022-10-31
Use byte as the unit for response size in all report formats

shelld3v:


maurosoria:
May I know the reason of this patch?

shelld3v:
> May I know the reason of this patch?

I just think we shouldn't use human-readable size as the unit for reports because some paths may have the same readable size but different responses because the gap between response sizes is too tiny only a few bytes.

It also makes sorting by response sizes in HTML report works intendedly (#1239)

PR #1241, opened 2022-11-07
Small patch

shelld3v:


maurosoria:
What is this about?

shelld3v:
Some users reported that SSL warnings still appear for them. Even though I don't know why and how, I did a quick search and found this:
https://stackoverflow.com/questions/48767143/how-to-suppress-warnings-about-lack-of-cert-verification-in-a-requests-https-cal

So I made this fix, not sure if it really solves any issue but why not?

PR #1280, opened 2023-02-13
Fix "almost same" pages' success

### LINKED ISSUE #1279: For 'almost' static pages returned found.
When I'm trying to scan not static page, which is returns almost similar page content, all results will be "200 found"

Dirsearch make flag DynamicContainerParser._is_static = True because of first test for page's diff showed that pages are similar. Somewhere after start of scan some id inside page changes, but because of flag is_static Dirsearch doesn't use diff with _differ_compare we got wrong results of page's difference

Compare pages with differ_compare if page's content doesn't equal to base_content in DynamicCompareScanner

image

kosyan62:
Description
---------------

What will it do?
It fixes bug when almost same pages on different endpoints return a lot of "found". 

Fix #1279 

Requirements
---------------

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -71,6 +71,7 @@
 - [at0m](https://github.com/atomiczsec/)
 - [junmoka](https://github.com/junmoka)
 - [Akshay Ravi](https://www.linkedin.com/in/c09yc47/)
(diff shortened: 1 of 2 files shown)


maurosoria:
Hello @kosyan62 ! Thanks for the fix!

maurosoria:
Hello @kosyan62 ! Thanks for the fix!

PR #1348, opened 2023-12-30
1341 Issue fix -  reports folder is created (empty) when you specify output flag

### LINKED ISSUE #1341: "reports" folder always created
### What is the problem?

The "reports" folder is created regardless of whether you specify a different path for the output file. Additionally you cannot set a flag for disabling output files.

### How should it be solved?

There should be a flag to disable output file and it should not create the "reports" folder on the current working directory if a different path is specified via the `-o` option.

ajcriado:
Description
---------------

The reports folder is always created where you execute the app even though you specify output flag because controller.setup method is not handling the options["output"] value, so it creates the folder ("reports" name by default, stored in options["output_path"]) leaving it empty.

I just control if options["output"] comes with value as it is done in controller.setup_reports method

Fix #1341

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -77,6 +77,7 @@
 - [huyphan](https://github.com/huyphan)
 - [Sean Wei](https://www.sean.taipei/about-en)
 - [FantasqueX](https://www.github.com/FantasqueX)
+- [ajcriado](https://www.github.com/ajcriado)
 
 Special thanks to all the people who are named here!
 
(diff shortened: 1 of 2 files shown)


maurosoria:
Hi Antonio! Thanks for the contribution!

Seems a valid fix @shelld3v, right?

shelld3v:
This should work, but I'm seeing a very similar check here.
image

I'm wondering whether we should move that piece of code (creating report folder for auto-save reports) to `setup_reports`? I don't even know why I didn't do it from the start, was it because of an unknown issue? 🤔

PR #1349, opened 2024-01-02
1305 Issue fix - delay applied to all requests

### LINKED ISSUE #1305: Baseline Requests do not respect "delay" flag
When using the `--delay` flag to avoid rate limits, the initial HTTP requests that are sent to establish what the baseline "not found" responses should be (one for each supplied extension) do not respect the delay and are sent in quick succession. I have to proxy `dirsearch` through BurpSuite, intercept the requests, and manually click continue on the first few requests to prevent being rate limited. After that it works fine.

All HTTP requests sent should respect the `--delay` flag, even the ones intended to tune detection. 

**Command:** `dirsearch -u https://REDACTED/ -e php,txt,html,md,xml,bk --delay=1 -t 1`

Notice the timestamps:
image

ajcriado:
Description
---------------

Fix #1305 Issue

Delay parameter is not applied to baseline requests so I have changed the Time.sleep call from fuzzer.py to requester.py, where all the requests are made

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -80,6 +80,7 @@
 - [Ovi3](https://github.com/Ovi3)
 - [u21h2](https://www.github.com/u21h2)
 - [Valentijn Scholten](https://www.github.com/valentijnscholten)
+- [ajcriado](https://www.github.com/ajcriado)
 
 Special thanks to all the people who are named here!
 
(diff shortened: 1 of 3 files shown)


maurosoria:
Hello Antonio!

IMO delay should be applied between baseline requests. In terms of responsibility, the Requester shouldn't be in charge of scheduling this because most of the time runs with multiple threads. If you want to put this part of code between baseline requests I'd be happy to merge it.

Thanks for contributing!

PR #1359, opened 2024-01-27
Remove ioerror

jxdv:
Description
---------------

Since dirsearch requires >= python 3.7 IOError can be removed since it's been deprecated in python 3.3 and merged into OSError.

Requirements
---------------


shelld3v (approved):
Good catch!

Issue #1357, opened 2024-01-23
Broken output folder name when scheme isn't provided


shelld3v:
### What is the current behavior?

When you provide a target without an URI scheme (`-u example.com`), the URI scheme will be missing in the output folder name, as below

![image](https://github.com/maurosoria/dirsearch/assets/59408894/50cbc370-2e40-4e9e-836f-b0c536da52d6)

### What is the expected behavior?

Because dirsearch detects the scheme anyway, the scheme should appear in the folder name, as below

![image](https://github.com/maurosoria/dirsearch/assets/59408894/fbe71539-6e53-478c-97f9-77e3de257cdb)

maurosoria:
This should be easy to fix. Moreover, was the older behaviour

shelld3v:
> This should be easy to fix. Moreover, was the older behaviour

The fix is probably performing scheme detection before the creation of the report folder, I will work on this when I have time

nrathaus:
I believe offending code is inside `def setup_reports(self):`

PR #1370, opened 2024-04-30
set_target before we parse the URL and then use self.url

nrathaus:
# Description

What will it do?

https://github.com/maurosoria/dirsearch/issues/1357

# Recreate issue

Without the fix run:
`dirsearch "--url=scanme.nmap.org"`

Report will be generated at:
`/reports/_scanme.nmap.org/__24-04-30_15-05-07.txt`

With fix run and the report will be stored at:
`/reports/http_scanme.nmap.org/__24-04-30_15-05-07.txt`

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -374,7 +374,9 @@ class Controller:
                 directory_path = self.setup_batch_reports()
                 filename = "BATCH." + self.get_output_extension()
             else:
-                parsed = urlparse(options["urls"][0])
+                self.set_target(options["urls"][0])
+
+                parsed = urlparse(self.url)
 
                 if not parsed.netloc:
                     parsed = urlparse(f"//{options['urls'][0]}")


shelld3v:
Won't this make us run `set_target` twice? One in `run()` and one in `setup_reports()`?

nrathaus:
I didn't observe this - are you seeing this happening?

shelld3v:
Luckily, running `set_target` twice doesn't cause any significant impact here, but it still makes us waste some HTTP requests, as well as a weird-looking code, therefore we still have to change it. What's your idea of fixing, @nrathaus?

shelld3v:
Also this won't fix the bug you mentioned in the description, if you give it a list of URLs. For example, a file containing targets like this:
```
https://example.com
scanme.nmap.org
```
It will still give us `_scanme.nmap.org` instead of `http_scanme.nmap.org`.

shelld3v:
@maurosoria Hi can you revert this PR real quick? It doesn't really address the issue and actually causes more issues

nrathaus:
The idea of the fix was to pass to `set_target` the scheme, as set_target created the file/folders

Issue #1313, opened 2023-07-28
Auto-Extract Domain in Dictionary


v1zr38g4m8op13azq3:
**I hope this message finds you well. First of all, I want to express my appreciation for dirsearch; it's a fantastic tool for web directory scanning.

I have been using dirsearch for some time now and have found it extremely useful in my security assessments. I would like to request a new feature that I believe would enhance the tool's flexibility even further.

The feature I'm proposing is to automatically extract the target's domain and main domain when %domain% is included in the dictionary. This would enable users to customize the scan for backup files more efficiently. For example, by adding %domain%.zip to the dictionary, the tool would be able to scan for files like test.test.com.zip and test.com.zip.

By implementing this functionality, dirsearch users would have a more streamlined way to scan for relevant backup files without the need to explicitly specify each subdomain variation.

I understand that you and your team have a lot on your plate, but if it's feasible, I believe this feature would greatly benefit the dirsearch community. I'm happy to provide any additional insights or collaborate in any way to help make this feature a reality.

Thank you for your time and consideration. I look forward to hearing from you.**

maurosoria:
hello (please introduce your name) how you doing?

Thanks for asking! actually we always wanted to add new keywords to the wordlist processor. What you say is something like this:
```
%DOMAIN%.tar
%BASE_DOMAIN%.zip
```
And the resulting processing from running dirsearch -u example.com
```
example.com.tar
example.zip
```

This is the intended functionality ?

v1zr38g4m8op13azq3:
Thank you for your response, Mauro Soria!

Yes, exactly! That's precisely the intended functionality I was suggesting. Being able to use %DOMAIN% and %BASE_DOMAIN% as placeholders in the wordlist processor would be extremely helpful for customizing the scan according to the target's domain and main domain.

Having this feature would make dirsearch even more versatile and efficient, allowing users to dynamically generate filenames specific to the target being scanned. It would save a lot of time and effort by automating the process of handling various subdomains and variations.

I truly appreciate your consideration in implementing this feature. If there's anything I can do to assist or provide further feedback, please feel free to let me know. I'm looking forward to seeing this enhancement in dirsearch!

Thank you again for your dedication to the project and for taking the time to discuss this suggestion.

Best regards,
Kevin li

v1zr38g4m8op13azq3:
please let me add
%domain%>>>%domain%.zip>>>https://www.example.com/www.example.com.zip and https://www.example.com/example.com.zip
[...]

PR #1371, opened 2024-04-30
Add code to support %DOMAIN% tag replacer

nrathaus:
# Description
What will it do?

https://github.com/maurosoria/dirsearch/issues/1313

if it finds a `%DOMAIN%` tag inside the dictionary, it will replace it with the `hostname` field of the URL - if multiple URLs are provided it will replace it multiple times

### DIFF
--- a/lib/core/dictionary.py
+++ b/lib/core/dictionary.py
@@ -144,7 +144,7 @@ class Dictionary:
                 if EXTENSION_TAG in line.lower():
                     extension_tag_triggered = True
 
-                if DOMAIN_TAG in line.lower():
+                if DOMAIN_TAG in line:
                     domain_tag_triggered = True
 
                 final_lines = []
--- a/lib/core/dictionary.py
+++ b/lib/core/dictionary.py
@@ -18,10 +18,13 @@
 
 import re
 
+import urllib.parse
+
 from lib.core.data import options
 from lib.core.decorators import locked
 from lib.core.settings import (
     SCRIPT_PATH,
+    DOMAIN_TAG,
     EXTENSION_TAG,
     EXCLUDE_OVERWRITE_EXTENSIONS,
     EXTENSION_RECOGNITION_REGEX,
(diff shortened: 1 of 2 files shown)


maurosoria:
Hi! Glad to meet you @nrathaus ! How you doing?

I like it but I wonder, for example if type www.example.com:
* Shouldn't try example.com?
* Or just example?

For example, given the case "%DOMAIN%.zip":
* www.example.com.zip
* example.com.zip
* example.zip

nrathaus:
@maurosoria I added the iterator so that it will go through the `components` of the domain

Let me know if this is what you were expecting

maurosoria:
Hi again!

I see you got me but there are 3 issues with your commit:
* What happens when a line has both %EXT% or %DOMAIN% in the same line (or any other future tag)? I see in this commit you only allow one tag per line
* domain can be parsed before entering the loop in line 125 (also you can use tldrextract or a regexp to parse the domain)
* Same behaviour should be applied at line 196


Thanks for your quick response. If you feel I am too demanding please tell me. I Like your patch (I also like other ones like %YEAR%!!!)!

nrathaus:
I did a one tag replacer because EXT had a similar design, to support multiple tags - we would need to rewrite the whole function used - I am ok with doing that

nrathaus:
Regarding `tldrextract`, I didn't want to introduce another library - we don't actually need this in my opinion, just dot breaking is sufficient for this `dirsearch` improvement

I added additional code, didn't get a chance to test it yet - will try to do it later

shelld3v:
Another problem I found in this PR: Assuming you run on 2 targets: `abc.com` and `xyz.net`, your code will basically create `abc.txt` and `xyz.txt` from `%DOMAIN%.txt` apply on BOTH targets, even though I only want to run `abc.txt` on `abc.com` and `xyz.txt` on `xyz.net`

maurosoria:
Hello @nrathaus, how you doing?

Do you mind contacting us ? We really want to introduce this PR with @shelld3v !

nrathaus:
Not sure how to do that

PR #1376, opened 2024-05-14
Add backup file adder code

nrathaus:
# Description
Add support such that when a file is found (200 HTTP OK code), additional extensions (related to backup files) are added to the queue

# TODO
Some functionality is done quite ugly, let me know how you want to support it
1. `backup extensions` are hardcoded to a list, should I read it from a file? a config parameter?
2. `_items` is directly accessed, should I add a `Dictionary` func to add items? is it allowed to add to `_items` running it appears to have no ill effect

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -442,6 +442,11 @@ class Controller:
 
         interface.status_report(response, options["full_url"])
 
+        if response.status == 200 and not response.full_path.endswith("/"):
+            backup_extensions = [".bak", "~", ".swp", ".swn", ".tar.gz", ".zip"]
+            for backup_extension in backup_extensions:
+                self.dictionary._items.append(f"{response.full_path}{backup_extension}")
+
         if response.status in options["recursion_status_codes"] and any(
             (
                 options["recursive"],


shelld3v:
Everything that's done behind the scenes must be known and decided by users, if you make a flag like `--discover-backups` I might consider it, even though I still don't think this is a good idea, because if you have found the file, why would you need the backup? And how many cases like this actually exist in the wild? Still appreciate it tho 🙏🏻

iseries:
`sed 's/$/.bak/' worlist.txt > bak_wordlist.txt`?

nrathaus:
The idea is to find a file that exists, then see if .bak version of it exists - not to duplicate the number of files queried for by adding another extension

shelld3v:
> > Everything that's done behind the scenes must be known and decided by users, if you make a flag like `--discover-backups` I might consider it, even though I still don't think this is a good idea, because if you have found the file, why would you need the backup? And how many cases like this actually exist in the wild? Still appreciate it tho 🙏🏻
> 
> when performing actual web page security checks, we often find that dynamic pages such as `index.php` and `conn.php` typically aren't parsed by middleware, which means we can't view their source code directly. If there is a backup version of these files, like `conn.php.bak`, we can successfully obtain the source code. Currently, we are unable to perform further detection on discovered files; instead, we simply append known file extensions to our dictionary, which is inefficient and results in a large number of HTTP requests being sent.

Would the source code contain anything more informative than the page itself tho (like something really sensitive)? And I would like to know how common this is, and how many people actually need this feature

PR #1383, opened 2024-07-31
Update dicc.txt

AkshayraviC09YC47:
Hello Team,

From our recent assessment, we have found that developers are using these "api/1, api/v1/1" paths and files to backup their source code and other internal files. These paths were not available in Dirsearch.

### DIFF
--- a/db/dicc.txt
+++ b/db/dicc.txt
@@ -1354,6 +1354,8 @@
 .zshrc
 .zuul.yaml
 .zuul.yml
+api/1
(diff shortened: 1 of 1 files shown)


shelld3v:
Just because it exists in one server doesn't mean it exists in most servers, I don't think those paths are worth it, so sorry for closing this. If you can show a third-party application/framework where a path like this exists, I will re-consider it. Appreciate your work tho 🙏

PR #1391, opened 2024-09-07
Support asynchronous mode

### LINKED ISSUE #643: Moving from threads to asyncio
Hello folks,

I am opening this issue because I do not have the time to do it myself, but I really think that now is the momentot move from threads to an asyncio based application (at least for requests and some callbacks).

I planned from the beginning to do this, but back then asyncio wasn't widely supported by most of linux distros and wasn't as mature like it is today. 

This is a very important step, because it would open the door to add extra scanning capabilities to dirsearch, that now seems very difficult with a normal thread scheme.

### What is the feature?

Moving from a regular thread-based system to an asynchronous based requests system would improve the performance.

### Take into consideration
* The idea is using an asyncio executor for processing Jobs.
* It is important to follow pep8 guidelines and validate the code.
* It might be a good idea considering a model who may, in the future, support scanning multiple hosts at the same time (like [meg](https://github.com/tomnomnom/meg))
* I guess the proper library for doing this may be [aiohttp](https://docs.aiohttp.org/en/stable/), but I guess for now it is important to stop using threads and keep using [requests](https://requests.readthedocs.io/en/master/).
* It can be very time consuming to code, but maybe using CPU bound executor for certain tasks could make a difference (maybe not).

Obviously, I will assist you if you want to cooperate in all I can, and I will review the code myself.

Thanks for cooperating, and spread the voice!

Regards, 
Mauro

zrquan:
Description
---------------

This PR adds the `--async` parameter to enable asynchronous mode (close #643). In asynchronous mode, coroutines are used instead of threads to send network requests (use `httpx`).

I ran a few quick tests and found that the asynchronous mode really speeds up the scanning process. (`uvloop` can makes it even faster)

```
python dirsearch.py -u https://security-crawl-maze.app -e html -t 50  150.96s user 5.89s system 98% cpu 2:38.45 total
python dirsearch.py -u https://security-crawl-maze.app -e html --async  41.34s user 0.94s system 40% cpu 1:43.80 total
python dirsearch.py -u https://security-crawl-maze.app -e html --async -t 50  33.49s user 0.65s system 53% cpu 1:03.24 total
```

**This PR isn't quite finished yet!** 
- [x] I just copied and modified the existing code, so it needs to be refactored.
- [x] The rate-limiting logic is likely to require further optimisation.
- [x] HttpNtlmAuth hasn't been implemented yet.
- [x] There might be some bugs in concurrency and timeout control.
- [x] <s>Should we replace requests with httpx? Since httpx supports both synchronous and asynchronous modes.</s> There should be a new issue for discussion.
- [x] Handle all request exceptions properly.
- [x] Change the minimum version requirement to Python3.9. (`httpx` requires 3.8+)
- [x] Adjust log/output in async mode.
- [ ] Fix errors from pickle (requester, loop, future...)
- [ ] `httpx` doesn't let me choose [different proxy for each request](https://github.com/encode/httpx/discussions/3183).

Requirements
---------------


zrquan:
Should we think about raising the minimum version requirement to Python3.9?

maurosoria:
I like this! 

We can jump to even newer versions if it is worth it.

DM me on twitter if you want to chat with us.

maurosoria:
Thank you so much and congratulations for such important contribution!!!!

PR #1394, opened 2024-09-13
add --urls-only option

zrquan:
Description
---------------

This PR added the `--urls-only` option to make dirsearch output only the `full_url`, which is useful to pipe the output to other tools.

Like: `dirsearch -u http://172.18.0.2:8080 -e html -i 200 --urls-only | httpx-toolkit -title`

Requirements
---------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Support non-default network interface
 - Remove unused dependencies (urllib3, cryptography, cffi, idna, chardet)
 - Load targets from a Nmap XML report
+- Added --silent option to output only the full URL
 
 ## [0.4.3] - October 2nd, 2022
 - Automatically detect the URI scheme (`http` or `https`) if no scheme is provided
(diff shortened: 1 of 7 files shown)


shelld3v:
I don't think `--silent` is a good name, probably `--urls-only` is what I would accept. However, I still don't know if this is worth an option, as there are already too many options and many more will be added in the future

zrquan:
> I don't think `--silent` is a good name, probably `--urls-only` is what I would accept

Okay, I used this name just because ProjectDiscovery tools use it.

> I still don't know if this is worth an option

From my own experience, this is more useful than quiet mode. I'd love to start a discussion to see what others think.

zrquan:
How about `-qq, --quite-quiet`, just like `-vv` in some tools 🤔

shelld3v:
I just realized `simple` output format actually gives you the same output as this. Why don't you use that? 🤔

zrquan:
> I just realized `simple` output format actually gives you the same output as this.

My bad—I didn't catch this parameter earlier.

PR #1405, opened 2024-10-13
update README and config.ini for async mode

zrquan:
Description
---------------

Add description and config about async mode.

Requirements
---------------

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ Options:
 
   Advanced Settings:
     --crawl             Crawl for new paths in responses
+    --async             Enable asynchronous mode
 
   View Settings:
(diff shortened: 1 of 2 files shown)


shelld3v:
Do you think **--async** should be considered a **General** feature instead of **Advanced**? That's the way I see it right now, it's more than just something "extra"

zrquan:
Yep, I think asynchronous is a better bet than multithreading for dirsearch. Once we get the "session saving" issue sorted, we can think about replacing the original multithreading with asynchronous execution.

shelld3v:
@zrquan If so would you like to move the option to **General** category? I will approve the PR once you submit

PR #1408, opened 2024-10-15
Options to disable all the CLI output

### LINKED ISSUE #1230: Option to disable CLI output
### What is the feature?

Option to disable the command-line output, users can only get the results through report files

shelld3v:
Description
---------------

Close #1230


zrquan (changes requested):
Update README?

shelld3v:
> Update README?

I think there are a lot of things to change in README, why not fix them all once before making the release? 😀

PR #1410, opened 2024-10-15
Do not add '/' when '-f' is used

shelld3v:
Description
---------------

Currently `-f` does not only append extensions to paths but "/" as well. This unnecessarily generates a huge number of entries, and users can't get rid of this behavior no matter what. So, I decided to remove it, considering that users still can append slash if they want with `--suffixes`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Wordlists (IMPORTANT)
 **Summary:**
   - Wordlist is a text file, each line is a path.
   - About extensions, unlike other tools, dirsearch only replaces the `%EXT%` keyword with extensions from **-e** flag.
-  - For wordlists without `%EXT%` (like [SecLists](https://github.com/danielmiessler/SecLists)), **-f | --force-extensions** switch is required to append extensions to every word in wordlist, as well as the `/`.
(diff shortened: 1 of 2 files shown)


maurosoria:
Hello! I added this because in the past  some users missed endpoints (because webapp routing systems are sensitive to a slash at the  end). This wasn't the case many years ago, but now the slash we'll probably find different endpoints.

It is true that not every web framework have this behaviour and I open to discussion, but I don't think this a solution.

shelld3v:
@maurosoria Maybe create something like `--slash`? Cause I would like this behavior to be "optional"

maurosoria:
If there is an option to change manually force this behaviour it would be much better. Anyway, I think keeping the current behaviour by default and then what we can do is  provide more curated and largest wordlists.

shelld3v:
"Providing more curated and larger wordlists" isn't the solution because our default wordlist doesn't even need `-f`, users use this on their own crafted wordlist, something that we can't control. And the problem here isn't much related to the wordlist anyway, the problem is that some web servers treat `/path` and `/path/` equally, so generating tens of thousands more entries for absolutely nothing is a massive waste

PR #1415, opened 2024-10-17
Update dicc.txt - logout.php

drego85:
Description
---------------

Added `logout.php`


shelld3v:
I think `logout.%EXT%` is enough to cover `logout.php`

PR #1430, opened 2024-11-01
use signal.signal() on Windows

### LINKED ISSUE #1423: Async mode doesn't work well on Windows

If you use async mode on Windows, you will get `NotImplementedError` due to `add_signal_handler`. Reference: https://stackoverflow.com/questions/45987985/asyncio-loops-add-signal-handler-in-windows

zrquan:
Fix #1423

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -205,7 +205,12 @@ class Controller:
         self.requester = Requester()
         if options["async_mode"]:
             self.loop = asyncio.new_event_loop()
-            self.loop.add_signal_handler(signal.SIGINT, self.handle_pause)
+            try:
+                self.loop.add_signal_handler(signal.SIGINT, self.handle_pause)
+            except NotImplementedError:
+                # Windows
+                signal.signal(signal.SIGINT, self.handle_pause)
+                signal.signal(signal.SIGTERM, self.handle_pause)
 
         while options["urls"]:
             url = options["urls"][0]


shelld3v:
@zrquan Hey are you sure this works well? Cause I thought back then that if we used `signal.signal` then all the exception handlers wouldn't receive the exception calls from `self.handle_pause` right? Even though it did work well in my tests but I still couldn't figure out why it worked and would there be any issue.

shelld3v:
And if we can make sure that `signal.signal` works perfectly well then I don't see why we should do the same to Linux and even the default "threads" mode

zrquan:
And I found a bug in "threads" mode: when skipping the current directory to the next one, the program just exits.


https://github.com/user-attachments/assets/27e7c937-dc93-4719-b990-ae98874f723b

shelld3v:
@zrquan Yeah, I think exceptions raised in `handle_pause` can't be catched in `start`, so probably by far we should only use `signal.signal` in async mode. Bringing it to threads mode will take more than just a few lines of code

zrquan:
> I think exceptions raised in `handle_pause` can't be catched in `start`

I tested it on Linux, and it can catch exceptions just fine.

The bug I mentioned earlier happens because the "next" operation [sets the fuzzer's quit_event](https://github.com/maurosoria/dirsearch/blob/f1a1d30a79f139600d0978a22c569d06e5a98991/lib/controller/controller.py#L501), but it isn't cleared in the next scan, which stops the task. I fixed it in commit 82eeb84.

PR #1436, opened 2024-11-08
Add option to filter duplicate results and deprecate --remove-extensions

### LINKED ISSUE #1293: Suggestions for a filter flag to improvie accuracy
### What is the feature?

The feature sets a flag to skip continuously getting the same size of response and exclude specific paths during recursive scans.

### What is the use case?

This feature is useful when scanning websites that have a web application firewall, redirects, or custom error pages that return continuously the same size of 200 response. By setting this flag, dirsearch can skip these continuously same size responses and exclude specific paths to produce more accurate results.

shelld3v:
Description
---------------

Close #1293


shelld3v:
Hi @mikhailevtikhov, thanks for your feedback! Yes, I'm aware of your problem already but storing and comparing anything more than a hash is just way too expensive, I don't want people to complain about memory and performance. An idea is to store just a part of the responses, and to find a method to identify potential duplicates before performing high-level comparison. However, I'm avoiding any big changes before the release of v0.4.5, so I won't work on it at the moment

maurosoria:
hello @shelld3v can u resolve and merge?

shelld3v:
@maurosoria Done 👍

maurosoria:
is this ready?

PR #1438, opened 2024-11-13
fix proxy problem

zrquan:
Fix the "invisible proxy support" problem.

⚠️ It's conflicting with #1437, so I'll take care of it once that's merged.


shelld3v:
Hey @zrquan, I can add this fix to #1437 if you don't mind

PR #1461, opened 2025-02-20
Exclude "/css/" from recursion

shelld3v:


maurosoria:
I dislike this. First, user should know at every moment that "css" directory is being skipped. Let's assume there is an interesting finding in css folder, the tool would miss those findings.

shelld3v:
The chance of having something interesting in the `/css/` folder is way too low that it's not worth of the resources and the time we spend on it at all. In case an user actually wants to check that because there is nothing else to check, he can do it manually

ordinary-hacker:
in my opinion it would be better to keep stuff as is, the user can simply use the `--exclude-subdirs` flag so I don't find a reason for this

PR #1474, opened 2025-06-07
Fix #1444: Replace deprecated pkg_resources with importlib.metadata

### LINKED ISSUE #1444: Unexpected SSL error
I am using Python 3.12.6 to run dirsearch to scan HTTPS websites, but I am getting an 'Unexpected SSL error'. I have tried updating the SSL certificates used by Python and reinstalled the Python certificate module 'certifi', but the issue persists.

CleanShot 2024-11-28 at 17 53 44@2x

jirpo9:
Description
---------------
This PR addresses issue #1444 by replacing the deprecated `pkg_resources` module with the modern `importlib.metadata` approach 

What will it do?
-----------------
- Replaces deprecated pkg_resources with importlib.metadata
- Adds custom exception classes to maintain compatibility
- Updates dependency checking to use modern Python packaging tools
- Ensures compatibility with Python 3.12+

Fix #1444

Requirements
---------------

- [jirpo9] Add your name to `CONTRIBUTORS.md`

### DIFF
--- a/dirsearch.py
+++ b/dirsearch.py
@@ -32,8 +32,9 @@ if sys.version_info < (3, 9):
     sys.exit(1)
 
 # silence pkg_resources deprecation warnings
-warnings.simplefilter("ignore", DeprecationWarning)
-from pkg_resources import DistributionNotFound, VersionConflict  # noqa: E402
+
+# Import custom exceptions instead of pkg_resources
+from lib.core.exceptions import DistributionNotFound, VersionConflict  # noqa: E402
 
 
 def main():
--- a/lib/core/exceptions.py
+++ b/lib/core/exceptions.py
@@ -51,3 +51,13 @@ class QuitInterrupt(Exception):
 
 class UnpicklingError(Exception):
     pass
+
+
+class DistributionNotFound(Exception):
+    """Exception raised when a required package is not found"""
+    pass
+
+
+class VersionConflict(Exception):
+    """Exception raised when there's a version conflict with dependencies"""
+    pass
--- a/lib/core/installation.py
+++ b/lib/core/installation.py
@@ -17,37 +17,66 @@
 #
 #  Author: Mauro Soria
 
-from __future__ import annotations
-
-import subprocess
 import sys
-import pkg_resources
+import subprocess
+from importlib import metadata
+from packaging import requirements, version
 
-from lib.core.exceptions import FailedDependenciesInstallation
-from lib.core.settings import SCRIPT_PATH
-from lib.utils.file import FileUtils
-
-REQUIREMENTS_FILE = f"{SCRIPT_PATH}/requirements.txt"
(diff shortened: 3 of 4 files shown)


shelld3v:
The funny thing is, I wrote that code to check whether users have installed all the required packages from `requirements.txt` and install it for them, but now that code itself requires an installation of a package from `requirements.txt`! That doesn't sound right

maurosoria:
Hi! Thanks for contributing @jirpo9 ! Welcome!

A few comments:

* The install_dependencies function was modified to drop stdout. Any reason for that? I assume it’s intentional, but I’d like to understand the motivation — we usually suppress stdout there to keep the CLI clean.
* The PR switches from using the REQUIREMENTS_FILE constant (f"{SCRIPT_PATH}/requirements.txt") to just "requirements.txt". That may break execution from outside the script’s directory. We rely on the absolute path to avoid issues when dirsearch is called from other locations.
* get_dependencies() is now parsing the requirements file manually, including handling # comments and empty lines. If we’re going to introduce a third-party module for this, maybe we should use it properly for parsing (e.g., with packaging or pip._internal.req), not half-manual parsing.

Lastly, in requirements.txt, libraries like packaging that are required for parsing should ideally go first — so they can be used before the rest of the file is processed. Otherwise we risk dependency checks failing midway due to missing modules.

jirpo9:
Hi @maurosoria and @shelld3v! You both make excellent points. After reconsidering, I realize my approach created more problems than it solved.
The chicken-and-egg problem with packaging is indeed a showstopper. Let me propose a different approach:
Option 1: Minimal change

Keep using pkg_resources in installation.py since it's part of setuptools (which is already required)
Only replace pkg_resources with importlib.metadata in the main dirsearch code where we can safely assume dependencies are installed
Add deprecation handling to suppress the warning

Option 2: Simple stdlib-only parsing
pythondef get_dependencies():
    deps = []
    try:
        with open("requirements.txt", "r") as f:
            for line in f:
                line = line.strip()
                if line and not line.startswith("#"):
                    # Simple parsing without external dependencies
                    dep = line.split("#")[0].strip()
                    if dep:
                        deps.append(dep)
        return deps
    except FileNotFoundError:
        pass
    return deps
Which approach would you prefer? Or do you see a better solution?
I'll also revert:

The stdout suppression in install_dependencies()
The absolute path change for REQUIREMENTS_FILE

Thanks for the thorough review - this is exactly why code reviews are valuable!

shelld3v:
Hi @jirpo9

> Only replace pkg_resources with importlib.metadata in the main dirsearch code where we can safely assume dependencies are installed

There is only one place, one time that `pkg_resources` is used and that's the case where it can't be replaced with `importlib.metadata`

> Option 2: Simple stdlib-only parsing
>
> ...

Your function only returns the list of dependencies and does not check whether it has been installed or not yet.

PR #1540, opened 2026-04-15
Added mailinator directory and sort dic

drego85:
Mailinator allows Developers and QA Testing teams to automatically test their Email and SMS workflows like 2FA verifications, sign-ups, and password resets with trillions of inboxes at your fingertips.


shelld3v:
@drego85 Eventhough that service is real, I don't think a path like that ever exists, appreciate your work though!

PR #1537, opened 2026-03-02
Replace `setup.py` with `pyproject.toml` and fix `SyntaxWarning` in `fuzzer.py`

dark-vpr:
Description
---

### What does this PR do?

1. **Migrate from `setup.py` to `pyproject.toml`**
   Replaces the legacy `setup.py` (which depended on the deprecated `pkg_resources` API) with a modern `pyproject.toml`. This fixes:
   - `UserWarning: pkg_resources is deprecated as an API` on startup (Fix #1481)
   - `ModuleNotFoundError: No module named 'lib.core.installation'` when installing via `pip install git+https://github.com/maurosoria/dirsearch` (Fix #1498)

2. **Fix `SyntaxWarning`: `break` in `finally` block in `fuzzer.py`**
   Removes the problematic `break` inside a `finally` block that triggered a `SyntaxWarning` on Python 3.8+, while preserving the original control flow.

Requirements
---

- [x] Add your name to `CONTRIBUTORS.md`
- [x] If this is a new feature, then please add some additional information about it to `CHANGELOG.md`

PR #1541, opened 2026-04-28
Replace setup.py with pyproject.toml

mohamedsolaiman:
Migrates from deprecated setup.py to modern pyproject.toml (PEP 621). Fixes broken pip install caused by removed lib.core.installation module. Closes #1537

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,11 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install codespell flake8 -r requirements.txt
+    - name: Packaging smoke test
+      run: |
+        python -m pip install .
+        dirsearch --version
+        python tests/check_packaged_install.py
     - name: Test
       run: |
         python3 dirsearch.py -w ./tests/static/wordlist.txt -u https://example.com -o "tmp_report.{extension}" -O json,xml,plain,csv,md,sqlite,html --force-recursive -R 3 --full-url -q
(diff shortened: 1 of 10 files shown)


maurosoria:
Validation and follow-up work completed.

Changes applied on the PR branch:
- `f4541da` Fix pyproject packaging and verify Python 3.9 install
- `9c75c64` Add `python -m dirsearch` entry point
- `80f1df4` Use the checked-in PyInstaller spec in the GitHub workflows

What changed:
- Restored a working flat-layout packaging shim in `setup.py` so built installs keep the expected `dirsearch/` package layout.
- Single-sourced the package version from `lib/core/settings.py`.
- Moved runtime dependency resolution into the build script and removed `setuptools` from runtime dependencies.
- Added `__main__.py` so `python -m dirsearch ...` works after install.
- Added `tests/check_packaged_install.py` and wired it into CI.
- Corrected Python 3.9-incompatible pins:
  - `mysql-connector-python==9.4.0`
  - `psycopg[binary]==3.2.13`
- Normalized the four PyInstaller workflows to build from `pyinstaller/dirsearch.spec` instead of maintaining separate inline PyInstaller command blocks.

Verification completed:
- Clean Docker install on Python 3.9 passed.
- Clean Docker install on Python 3.14.4 passed.
- `dirsearch --version` passed after install.
- `python -m dirsearch --version` passed after install.
- Packaged-install verification script passed in both Docker checks.

Repository branch copy:
- Pushed the same fix series to `maurosoria/dirsearch:pr-1541-fix` for reference.

Note on workflow dispatch:
- I attempted to dispatch the Linux PyInstaller workflow on the contributor fork branch, but GitHub returned `404` because that workflow is not present on the fork’s default branch. The workflow definitions are now aligned with the checked-in spec, but I did not complete a remote Actions artifact download from the fork for that reason.

PR #1552, opened 2026-05-29
Use stealth words for wildcard calibration

maurosoria:
## Summary

Replaces high-entropy wildcard calibration tokens with deterministic pseudo-word chains and fixes an async wildcard parity issue.

## Changes

- Adds `StealthWordGenerator` with a precomputed pseudo-word bank, deterministic seed support, URL-safe separators, length constraints, and low-entropy filtering.
- Uses stealth words for sync and async wildcard calibration paths instead of short random alphanumeric strings.
- Keeps `rand_string()` for existing callers.
- Fixes `AsyncFuzzer` so it no longer adds an extra index-page wildcard scanner that sync/native do not use.
- Adds regression tests for entropy, format constraints, deterministic output, common-directory collision avoidance, omitted values, and the async index-page false negative.

## Validation

- `/home/mauro/dirsearch/.venv/bin/python -m unittest tests.utils.test_random tests.core.test_scanner tests.core.test_async_fuzzer`
- `/home/mauro/dirsearch/.venv/bin/python -m unittest discover tests`
- `/home/mauro/dirsearch/.venv/bin/python testing.py`
- `cargo test --manifest-path native/Cargo.toml`
- `git diff --check`

Note: `test_schemedet` still emits existing ResourceWarnings for unclosed SSL sockets; the suite passes.

### DIFF
--- a/lib/core/fuzzer.py
+++ b/lib/core/fuzzer.py
@@ -389,15 +389,8 @@ class AsyncFuzzer(BaseFuzzer):
 
     async def setup_scanners(self) -> None:
         # Default scanners (wildcard testers)
-        self.scanners["default"].update(
-            {
-                "index": await AsyncScanner.create(
-                    self._requester, path=self._base_path
-                ),
-                "random": await AsyncScanner.create(
-                    self._requester, path=self._base_path + WILDCARD_TEST_POINT_MARKER
-                ),
-            }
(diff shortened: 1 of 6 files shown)


hemu8980:
Ok i will work with you on anything but tomorrow I need 250$ urgent if you
can pay ***@***.*** is my wise account so please help me

PR #1560, opened 2026-06-01
Handle absolute-form targets in raw requests

xujiantop-crypto:
## Summary
- handle absolute-form request targets in raw request files, such as `GET http://example.com/admin HTTP/1.1`
- preserve query strings while using the absolute-form authority instead of prepending the `Host` header
- add raw request parser coverage and include it in the legacy `testing.py` runner

## Validation
- `python -m unittest tests.parse.test_rawrequest tests.parse.test_url tests.parse.test_headers -v`
- `python -m flake8 lib/parse/rawrequest.py tests/parse/test_rawrequest.py testing.py`
- `python testing.py` was also run locally; the raw request tests passed, but this Windows/Python 3.14 run still reports an unrelated async requester path-preservation failure where the local test server receives no async requests.

### DIFF
--- a/lib/parse/rawrequest.py
+++ b/lib/parse/rawrequest.py
@@ -18,10 +18,23 @@
 
 from __future__ import annotations
 
+from urllib.parse import urlsplit
+
 from lib.core.exceptions import InvalidRawRequest
 from lib.core.logger import logger
 from lib.parse.headers import HeadersParser
 from lib.utils.file import File
+
+
+def _target_from_request_line(path: str, host: str) -> str:
+    parsed = urlsplit(path)
+    if parsed.scheme and parsed.netloc:
+        path = parsed.path or "/"
+        if parsed.query:
+            path = f"{path}?{parsed.query}"
+        host = parsed.netloc
+
+    return host + path
 
 
 def parse_raw(raw_file: str) -> tuple[list[str], str, dict[str, str], str | None]:
@@ -47,4 +60,4 @@ def parse_raw(raw_file: str) -> tuple[list[str], str, dict[str, str], str | None
         logger.exception(e)
         raise InvalidRawRequest("The raw request is formatively invalid")
 
-    return [host + path], method, dict(headers), body
+    return [_target_from_request_line(path, host)], method, dict(headers), body
--- a/testing.py
+++ b/testing.py
@@ -52,6 +52,7 @@ from tests.core.test_scanner import TestScanner  # noqa: F401
 from tests.core.test_wordlist_backend import TestWordlistBackend  # noqa: F401
 from tests.parse.test_config import TestConfigParser  # noqa: F401
 from tests.parse.test_headers import TestHeadersParser  # noqa: F401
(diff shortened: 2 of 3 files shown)


maurosoria:
Thanks for the raw absolute-form request fix idea. I opened #1561 as a broader follow-up that builds on this PR: it keeps the absolute-form target handling, makes raw parsing byte-safe, adds Python API support through `FuzzerConfig.from_raw_request(...)`, preserves query/body behavior across sync and async requesters, validates raw GET/query through the native-rust stack, and credits you in `CONTRIBUTORS.md`.

maurosoria:
Thanks for opening this. The absolute-form raw request case is valid, and the current parser does mishandle it.

I opened #1561 as a broader follow-up based on your idea. It keeps the fix for requests like:

```http
GET http://example.com/admin?debug=true HTTP/1.1
Host: proxy.local
```

but extends it further:

- preserves the absolute-form scheme, host, port, path, and query string
- parses raw requests as bytes so request bodies are not corrupted by text decoding
- supports raw requests through the Python API with `FuzzerConfig.from_raw_request(...)`
- validates behavior across the current runtime stacks: `threaded`, `async`, and `native-rust`
- adds you to `CONTRIBUTORS.md`

You can test it from CLI like this:

```sh
python3 dirsearch.py --raw request.raw -w wordlist.txt -q
```

The default stack is async. You can force the threaded Python stack with:

```sh
python3 dirsearch.py --raw request.raw -w wordlist.txt -q --sync
```

And you can test the native Rust backend with:

```sh
python3 dirsearch.py --raw request.raw -w wordlist.txt -q --request-backend native
```

One important detail: the native backend currently supports only GET requests without request bodies. So raw GET requests should work there, but raw POST/body requests are intentionally rejected with an explicit message. For POST/body raw requests, test with the async default or `--sync`.

For Python API usage, the new flow is:

```python
from dirsearch import DirsearchFuzzer, FuzzerConfig

raw_request = (
    b"GET /admin?debug=true HTTP/1.1\r\n"
    b"Host: example.com\r\n"
    b"\r\n"
)

config = FuzzerConfig.from_raw_request(
    raw_request=raw_request,
    scheme="https",
    wordlist=["login", "dashboard"],
)

results = DirsearchFuzzer(config).run()
```

It also supports raw request files:

```python
config = FuzzerConfig.from_raw_request(
    raw_file="request.raw",
    scheme="https",
    wordlist=["admin", "login"],
)
```

For absolute-form request targets, the scheme and host from the request line are used directly. For origin-form targets like `GET /admin HTTP/1.1`, the API uses the `Host` header plus the provided `scheme`.

It would be very helpful if you could test #1561 from both the CLI and the Python API. If you can think of more edge cases for raw request parsing, especially around encodings, unusual request targets, or proxy-style requests, I would really appreciate them.

Thanks for everything,
Mauro Soria

xujiantop-crypto:
Thanks for taking this further in #1561 and for the detailed follow-up. Since that PR supersedes this narrower patch, I am closing this one to keep review focused there.

PR #1562, opened 2026-06-02
fix: replace empty/generic Exception with descriptive RuntimeError/ValueError

mshzy:
Fix 4 instances of bare raise Exception (2 with no message at all). Replace with descriptive RuntimeError and ValueError.


maurosoria:
Thanks for starting this cleanup. I agree there is value in avoiding bare/generic `Exception`, but the replacement messages should describe the actual failure domain.

I opened #1563 as the follow-up PR for this: it keeps the same goal, replaces the generic raises with more accurate built-in exception types/messages, narrows the obvious wrapper catches, preserves causes with exception chaining, and adds focused regression tests.

PR #1564, opened 2026-06-02
fix: improve exception handling with domain-specific types (closes #1563)

mshzy:
fix: improve exception handling with domain-specific types (closes #1563)

Replace bare/generic exceptions with domain-appropriate built-in types
and add descriptive error messages:

- controller.py: raise OSError with log file path on write failure
- csv_report.py: raise ValueError with expected vs actual CSV header
- sqlite_report.py: raise sqlite3.DatabaseError with exception chaining (from e)
- Preserve defensive broad `except Exception` at runtime fail-safe boundaries
- Add regression tests for CSV header validation and SQLite integrity check

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -192,7 +192,9 @@ class Controller:
                 try:
                     FileUtils.create_dir(FileUtils.parent(options["log_file"]))
                     if not FileUtils.can_write(options["log_file"]):
-                        raise Exception
+                        raise OSError(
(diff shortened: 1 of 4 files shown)


maurosoria:
Thanks for the PR and for the improvement around clearer exception messages.

This work is being superseded by #1563 so we can keep the exception-handling changes consolidated in one PR. I incorporated the useful CSV header mismatch detail there, and I also added you to `CONTRIBUTORS.md` so your contribution is credited.

Closing this PR in favor of #1563.

PR #1575, opened 2026-06-08
Add fingerprint matcher core

maurosoria:
## Summary
- Add `lib/core/fingerprint.py` with HTTP, cookie, body, TLS, CNAME, and IP/CIDR fingerprint matching.
- Keep CNAME/IP data injectable through a small `dataset=` argument instead of reading the large fingerprint JSON directly.
- Add focused tests with compact fixtures for header/cookie/body/TLS/CNAME/IP matching, negative benign headers, malformed dataset handling, provider scoring, behavior tags, and merge behavior.
- Register the new test module in `testing.py`.

## Why
This is PR4 in the phased rollout. It lands the matcher logic separately from the large CDN/WAF dataset so the review stays focused on behavior and scoring. The large `db/fingerprints/cdn_waf.json` file and attribution/licensing docs are intentionally left for the next PR.

## Validation
- `python -m unittest tests.core.test_fingerprint`
- `python testing.py` passed 149 tests with 2 skips.
- `python dirsearch.py -u https://example.com -w tests/static/wordlist.txt -q`
- `python -m flake8 lib/core/fingerprint.py tests/core/test_fingerprint.py testing.py`
- `git diff --check`


maurosoria:
Superseded by #1578, which keeps the same detection coverage but replaces the float-based confidence model with an explicit rule-based one (discrete per-signal `Strength` tiers + a rule-derived `Confidence` enum, no magic floats). Closing in favor of #1578.

PR #1599, opened 2026-08-20
Add finding backups feature

### LINKED ISSUE #1375: Add backup file auto detect
is it possible to add a `backup file` auto detect?
for example:
dirsearch **found** test.php
then search:
+ test.php.bak
+ test.bak
+ test.php~
+ test.php.swp
+ test.php.swn
+ test.php.tar.gz
+ test.php.zip...

using suffix / prefix will add fix to all content in dict, but i want dirsearch found file then add fix to it (not add to all content in dict)

shelld3v:
Description
---------------

Closes #1375

Requirements
---------------

- [x] PR text, comments, and committed files do not disclose private infrastructure details such as internal hostnames, machine names, SSH targets, IPs, regions, private paths, runner names, or operational topology


maurosoria:
@shelld3v, I updated this branch and would appreciate your review of the revised implementation.

The update:

- merges the current `master` branch and resolves the previous conflict;
- generates both full-name and basename variants (`test.php.bak`, `test.bak`, and `test.php~`);
- preserves the existing `%ARCHIVE%` / `%ARCHIVE_EXT%` placeholders while adding `%BACKUP%` / `%BACKUP_EXT%`;
- avoids recursively expanding media files and existing backups, using case-insensitive extension checks with a dot boundary;
- honors `--exclude-extensions` for dynamically queued candidates;
- restores unrelated wordlist, target fixture, documentation, and overwrite-extension changes;
- adds regression coverage for generation, CLI/config parsing, and dynamic queue consumption by the threaded, async, and native engines.

Local validation passed on Python 3.12 (325 tests, 18 optional skips) and Python 3.14 with the native extension available (325 tests).

Could you please check the updated code and confirm that the generated backup candidates match the behavior you intended for this feature?

shelld3v:
@maurosoria everything is good, however I have some feedback:
- Please change back to `testasp.vulnweb.com` because `testphp.vulnweb.com` has been taken down
- Remove the content of `config.ini` from the docs and instead add a link to the `config.ini` file itself
- `is_valid` check should be included in `Dictionary.add_extra`
- (Unnecessary) Consider removing `\..\..\..\etc\passwd` from the wordlist because it has been so annoying whenever I scan a target

shelld3v:
Btw, eventhough `a.php` to `a.zip` is not a "bad" thing to do, but `a.php` to `a.swp` or `a.swn` is really nonsensical. I suggest that we should do `a.php.%BACKUP%` + `a.%ARCHIVE%` instead

maurosoria:
@shelld3v Thanks for the detailed follow-up. I applied the feedback in two separate commits:

- `50c209a` centralizes dynamic-path validation in `Dictionary.add_extra()` and adds a direct regression test, limits basename candidates to archive suffixes (`test.php.swp`, but not `test.swp`), switches the integration target to `testasp.vulnweb.com`, and replaces the duplicated configuration block with a link to `config.ini`.
- `23a727c` moves the exact backslash traversal payload discussed here out of the default/common wordlists into the opt-in `aggressive` category. It also adds CLI, placement, and packaged-install regression coverage.

I intentionally left the other, distinct `/etc/passwd` discovery entries unchanged; this commit only relocates the noisy payload raised in the review.

Validation completed: 326 Python tests, 22 Rust tests, flake8, codespell, and a clean packaged-install check. Could you please take another look at the updated PR?

shelld3v:
@maurosoria Everything is good, you can merge it now 👍

PR #1603, opened 2026-08-22
Unify fuzzer response fingerprint bookkeeping

### LINKED ISSUE #1600: Duplicated work in Fuzzer
In `lib/core/fuzzer.py`, `self._hashes` and `self._auto_calibrated_fingerprints` are doing duplicated work, we should merge them into 1

fthsrbst:
`Fuzzer` tracked dynamic-page fingerprints in two separate structures (`_hashes` and `_auto_calibrated_fingerprints`) with overlapping bookkeeping — flagged in #1600. This unifies them behind a single fingerprint cache while preserving observable behavior exactly: calibrated fingerprints still prevent false-positive recursion on dynamic pages, static-page hash skipping still works, and filter ordering is unchanged.

`lib/core/fuzzer.py` holds the unified cache; `tests/core/test_advanced_filters.py` gains coverage pinning both legacy behaviors through the new path. Full suite: 193 passed, 4 skipped (pre-existing skips).

Fixes #1600

### DIFF
--- a/lib/core/fuzzer.py
+++ b/lib/core/fuzzer.py
@@ -48,6 +48,32 @@ AUTO_CALIBRATION_FORCED_THRESHOLD = 3
 AUTO_CALIBRATION_MIN_CONTENT_LENGTH = 32
 
 
+class FingerprintCache:
+    """Single bookkeeping cache for repeated-response fingerprints.
+
+    Tracks how many times each fingerprint was recorded and which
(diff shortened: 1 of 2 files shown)


shelld3v:
@fthsrbst can you check my comment: https://github.com/maurosoria/dirsearch/issues/1600#issuecomment-5384560125

PR #1614, opened 2026-08-24
Add --save-response to save found response bodies to a directory

### LINKED ISSUE #1441: Save responses in output-file
### What is the feature?

Hi! I suggest adding the ability to save the response from the server to the report.

### What is the use case?

This can be useful when analyzing results, especially when using dirsearch in automated processes.

ChrisJr404:
## Summary

Closes #1441.

Adds a `--save-response DIR` option that writes the raw response body of every found path into DIR, one file per hit. Handy for grepping through bodies after a scan or feeding them into other tooling in automated runs.

- filename is built from host, path, query and status code, sanitized so it stays valid on Windows too
- if two URLs sanitize to the same name a numeric suffix is added so nothing gets overwritten
- bodies are written as bytes, so binary responses stay intact
- the directory is created up front and the scan stops early with a clear error if it is not writable

It behaves the same across the sync, async and native backends since they all populate `response.body`.

## Validation

- `python -m unittest tests.utils.test_common tests.parse.test_cmdline tests.core.test_options`
- `python -m unittest discover -s tests -t .` reports 255 passed, 17 skipped
- manual run against a local server confirmed one file per found path with the expected bodies


maurosoria:
Thanks @ChrisJr404 for the original implementation and for getting this feature moving.

We have now merged the response-saving feature in #1615. It keeps the `--save-response` behavior proposed here and extends it with JSONL storage, shared store abstractions, filesystem and concurrency safeguards, full threaded/async/native parity, packaged-import coverage, and binary/multilingual response-body tests.

Your contribution is also credited in `CONTRIBUTORS.md` in the merged change. I am closing this PR as superseded by #1615. Thanks again!

PR #1616, opened 2026-08-26
Use per-dictionary locks and add concurrency tests

maurosoria:
## Summary

- replace the process-global `@locked` mutex used by `Dictionary` with a lock owned by each dictionary instance;
- make claiming, releasing, dynamic-path deduplication, state snapshots, and reset atomic for one dictionary;
- recreate the instance lock when dictionary state is restored;
- add deterministic concurrency coverage for independent dictionaries, duplicate dynamic paths, concurrent producers/consumers, and session snapshots with active claims.

## Motivation

`Dictionary` operations previously shared the same module-level lock used by unrelated dictionaries and other subsystems. A blocked operation on one dictionary therefore prevented another dictionary from making progress. `add_extra()`, `reset()`, and `__getstate__()` also mutated or copied queue state outside that lock.

The regression test first reproduced the issue by blocking one dictionary inside `claim_next()` and showing that a second independent dictionary could not claim its own item. With the instance-owned lock, the second dictionary completes independently.

This PR intentionally does not change pause handling, report locking, or HTTP test infrastructure; those remain separate follow-ups.

## Validation

- `python -m unittest tests.core.test_dictionary_concurrency tests.core.test_async_fuzzer tests.core.test_native_fuzzer`
- Python 3.12: 314 tests passed, 18 optional skips
- Python 3.14 with native extension: 314 tests passed
- `flake8 .`
- `codespell -S CONTRIBUTORS.md,./db/categories/*,./build`

### DIFF
--- a/lib/controller/controller.py
+++ b/lib/controller/controller.py
@@ -58,6 +58,8 @@ from lib.core.settings import (
     EXTENSION_RECOGNITION_REGEX,
     MAX_CONSECUTIVE_REQUEST_ERRORS,
     NEW_LINE,
+    PAUSE_POLL_INTERVAL,
+    PAUSE_TIMEOUT_SECONDS,
     SIGINT_FORCE_QUIT_THRESHOLD,
     SIGINT_WINDOW_SECONDS,
     STANDARD_PORTS,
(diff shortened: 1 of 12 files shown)


maurosoria:
Merge dependency: do not merge this PR yet. It depends on #1617. Merge #1617 into dictionary-concurrency first, then rerun the complete #1616 CI on the updated branch before considering the merge to master.

maurosoria:
Dependency satisfied: #1617 was merged into dictionary-concurrency as 7107e9e. The complete fresh CI on this updated head is green, including Python 3.11/3.14 on Linux and Windows, the native backend, all three Docker stacks, CodeQL, and Semgrep. This PR is now technically ready, but master protection still requires an approving review; no bypass was attempted.

PR #1647, opened 2026-09-10
Preserve threaded in-flight paths in sessions

maurosoria:
## Summary

- make threaded workers claim dictionary paths before starting a request
- release each claim only after request handling and callbacks finish
- preserve in-flight threaded paths in session snapshots, matching async and native behavior
- verify resume boundaries through the next queued directory and the next target on every stack

Previously, the threaded stack advanced the dictionary with `next()` before processing. If a session snapshot was written while that request was still in flight, resume started at the following word and silently skipped the unfinished path. `Dictionary` already provides claim tracking and includes outstanding claims in serialized state; this change adopts that contract in the threaded worker.

Session recovery now has the same at-least-once rule in every stack:

| Stack | Current interrupted job | Next directory/job | Next target |
| --- | --- | --- | --- |
| Threaded | Retry in-flight and continue pending paths | Full wordlist | Full wordlist |
| Async | Retry in-flight and continue pending paths | Full wordlist | Full wordlist |
| Native | Retry unreturned chunk paths and continue pending paths | Full wordlist | Full wordlist |

A path released after completed processing is not retried in the interrupted job. If completion races the snapshot, retrying is preferred to losing the path. Later directories and targets are independent jobs, so they intentionally start from the full wordlist.

## Tests

- the threaded in-flight regression fails against the base commit because `blocked` disappears and only `later` resumes
- stack-specific threaded, async, and native tests verify a fresh full wordlist after the resume state is reset
- a controller-level harness verifies the exact queue transition on all three stacks: current directory gets `in-flight` plus `later`; the next directory and next target each get `done`, `in-flight`, and `later`
- a JSON session round-trip verifies that queued directories, `jobs_processed`, remaining target URLs, and the partial dictionary cursor survive serialization
- `python -m unittest tests.core.test_fuzzer_lifecycle tests.core.test_async_fuzzer tests.core.test_native_fuzzer tests.controller.test_session_store tests.controller.test_session_resume_queue tests.controller.test_threaded_controller tests.controller.test_async_controller tests.controller.test_native_controller` (37 tests)
- `python -m unittest discover -s tests -t .` (437 tests, 20 skipped)
- `python -m compileall -q lib/core/fuzzer.py tests/core/test_fuzzer_lifecycle.py tests/core/test_async_fuzzer.py tests/core/test_native_fuzzer.py tests/controller/test_session_store.py tests/controller/test_session_resume_queue.py`
- `python -m flake8 lib/core/fuzzer.py tests/core/test_fuzzer_lifecycle.py tests/core/test_async_fuzzer.py tests/core/test_native_fuzzer.py tests/controller/test_session_store.py tests/controller/test_session_resume_queue.py`
- `git diff --check`

### DIFF
--- a/lib/core/fuzzer.py
+++ b/lib/core/fuzzer.py
@@ -474,8 +474,11 @@ class Fuzzer(BaseFuzzer):
         while True:
             should_quit = False
             try:
-                path = next(self._dictionary)
-                self.scan(self._base_path + path)
+                path = self._dictionary.claim_next()
+                try:
+                    self.scan(self._base_path + path)
+                finally:
+                    self._dictionary.release_claim(path)
 
             except StopIteration:
                 break
--- a/tests/controller/test_session_resume_queue.py
+++ b/tests/controller/test_session_resume_queue.py
@@ -1,0 +1,135 @@
+import asyncio
+from unittest import TestCase
+from unittest.mock import Mock, patch
+
+from lib.controller.controller import Controller
+from lib.core.data import options
+from lib.core.dictionary import Dictionary
+
+
+class RecordingFuzzer:
+    def __init__(self, _requester, dictionary, records, *args, **kwargs):
+        self.dictionary = dictionary
+        self.records = records
+        self.base_path = ""
+
+    def set_base_path(self, path):
+        self.base_path = path
+
+    def _record_job(self):
+        paths = []
+        while True:
+            try:
+                paths.append(next(self.dictionary))
+            except StopIteration:
+                break
+        self.records.append((self.base_path, paths))
+
+    def start(self):
+        self._record_job()
+
+    def is_finished(self):
+        return True
+
+    def stop(self, _timeout):
+        return True
+
+
+class RecordingAsyncFuzzer(RecordingFuzzer):
+    async def start(self):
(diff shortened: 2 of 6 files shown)


maurosoria:
Post-merge performance note

The correctness change keeps session recovery at-least-once, but the claim ledger remains Python-side. The Rust wordlist backend generates and expands entries, then PyO3 returns a Python `list[str]`; the shared Python `Dictionary` still owns its cursor, dynamic additions, reset, claims, and serialized session state.

The PR newly adds claim/release only to the threaded stack. Async and native already used that contract. A local isolated Python 3.11 microbenchmark measured approximately:

- raw `next()`: 190k paths/s
- uncontended claim + release: 101k paths/s
- native-style in-order 1,200-path chunks: 105k paths/s
- a deliberately no-I/O 12-thread contention case: 10k paths/s

These are data-structure measurements, not scanner-throughput claims. They do not show a current end-to-end bottleneck relative to the published full-scan results, but they justify a separate benchmark-gated optimization: batch claim/commit operations for native chunks and an O(1) ordered claim ledger, while preserving in-flight recovery. Removing claims would reintroduce silent path loss.

PR #1688, opened 2026-09-22
Fix async transport shadowed by httpx default mounts

### LINKED ISSUE #1687: Async backend returns zero results: all:// mount shadowed by httpx

The default async backend silently returns zero results on every target, without
printing any error and with exit code 0. The very same scan with `--sync` works
correctly and finds paths.

Reproduced against a local server that definitely serves known paths:

```console
$ python dirsearch.py -u http://127.0.0.1:18401 -w words.txt --no-color

  _|. _ _  _  _  _ _|_    v0.5.0
 (_||| _) (/_(_|| (_| )

Extensions: php, asp, aspx, jsp, html, htm | HTTP method: GET | Threads: 25
Wordlist size: 2

Target: http://127.0.0.1:18401/

[11:49:06] Scanning:

Task Completed
```

Note what happens after `Scanning:` — nothing. No result is printed, no error is
printed, and the process exits 0.

**Sync — same target, same wordlist, finds the paths:**

```console
$ python dirsearch.py -u http://127.0.0.1:18401 -w words.txt --sync --no-color

[11:49:12] Scanning:
[11:49:12] 200 -     6B - /index.html

Task Completed
```

This is not a target-side problem. The async backend never reaches the target
correctly because its custom transport is bypassed (see root cause below).

The default (async) backend should find the same paths as `--sync`. Since async is
the default backend, users who do not pass `--sync` currently get **silent false
negatives** — the user believes the target has no exposed paths.

- **dirsearch version**: v0.5.0, master @ `2b5c025` (2026-09-17)

- **httpx version**: `0.28.1` (pinned in `requirements/runtime.txt`)

- **Python**: 3.12.14, Windows 11

- **Command used**:

  ```
  python dirsearch.py -u http://127.0.0.1:18401 -w words.txt
  ```

- **No error is printed and the exit code is 0** — this is what makes the bug
  dangerous. The scan exits normally with zero findings, so the user concludes the
  target has no exposed paths.

**Root cause**

`lib/connection/requester.py`, in `AsyncRequester.__init__`:

```python
transport_options = (
    {"transport": transport}
    if options["proxies"]
    else {"mounts": {"all://": transport}}
)
```

With `httpx==0.28.1`, `httpx.AsyncClient` **pre-registers `http://` and `https://`
mounts internally**. `httpx` resolves a transport via `_transport_for_url()`, which
iterates over the mounts and returns the **first pattern that matches**. Because the
built-in concrete-scheme patterns are matched before the catch-all `all://`, the
custom `ScopedDNSAsyncTransport` is **never selected** and the request falls through
to the default transport.
**Fix**

Mount the concrete schemes explicitly instead of `all://`:

```python
else {"mounts": {"http://": transport, "https://": transport}}
```

Fix: mount the concrete schemes instead:

    else {"mounts": {"http://": transport, "https://": transport}}

Verified: with this change async and --sync return identical results.

Related: PR #1685 touches the same file but does not fix this — its branch still
uses {"mounts": {"all://": transport}}.

I can open a PR with a regression test if you think this is the right direction.

Cakefreeze:
Fixes #1687

## User-visible effect

The default async backend silently returned zero results on every target: the
custom `ScopedDNSAsyncTransport` was shadowed by httpx's pre-registered
`http://` / `https://` mounts, so the bare `all://` mount was never selected.
Users who did not pass `--sync` got silent false negatives with no error output
and exit code 0.

Mounting the concrete schemes keeps the scoped transport in the request path.
Async and `--sync` now return identical results.

## Root cause

`httpx.AsyncClient` pre-registers concrete scheme mounts and
`_transport_for_url()` returns the first matching pattern, so
`{"mounts": {"all://": transport}}` is never selected.

## Changes

- `lib/connection/requester.py`: mount `http://` and `https://` explicitly
- `tests/connection/test_requester.py`: add a regression test asserting that
  direct (no-proxy) requests resolve to `ScopedDNSAsyncTransport` for both
  schemes. The test fails on the previous `all://` mount and passes with this
  change.

## Commands run

python -m unittest tests.connection.test_requester
python -m unittest tests.connection.test_requester tests.connection.test_dns tests.core.test_scanner
python -m flake8 lib/connection/requester.py tests/connection/test_requester.py


maurosoria:
Thanks for the contribution. Closing this for now because the fix is incomplete and its regression test depends on the host proxy configuration.

With httpx 0.28.1, the original mount selects the custom transport in a clean environment. The proposed scheme mounts address shadowing by inherited HTTP/HTTPS proxies, but destinations covered by `NO_PROXY` can still bypass the custom transport. The new test passes against the original implementation when no proxies are configured.

I posted the transport-selection results and follow-up requirements in #1687, which remains open. A revised fix should cover inherited proxies and `NO_PROXY` with controlled tests and verify the reported scan behavior end to end.

Bathokoro221 (approved):

Bathokoro221 (approved):

